PR #8, opened 2018-11-21
Fix linux build_examples script to explicitly use bash.

atsushieno:
Because, the script actually depends on it.
    
Here's what happens on Ubuntu (dash):
    
```
tests/linux/build_examples: 36: tests/linux/build_examples: [[: not found
```

### DIFF
--- a/tests/linux/build_examples
+++ b/tests/linux/build_examples
@@ -1,4 +1,4 @@
-#!/bin/sh -e
+#!/bin/bash -e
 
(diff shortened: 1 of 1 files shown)


drowaudio:
Thanks. I wonder why this wasn't being marked as a failure in Travis though?

atsushieno:
Thanks for the (bloader) fixes.

Looks like travis logs are not publicly available, so I have no idea.

PR #10, opened 2018-12-14
Avoid double brackets in /bin/sh scripts

piem:


piem:
hi,

I hadn't seen #8, and doesn't see a commit referencing it, so i guess this requires an explanation:
 - `[[ test ]]` is a [bashism](https://mywiki.wooledge.org/Bashism), which assumes `/bin/bash` is being run, which is not [necessarily](https://wiki.ubuntu.com/DashAsBinSh) the case.
 - variables should be escaped with double quotes when using `[ ]`, which was already done here.

best, piem

atsushieno:
I had been seeing this kind of bash-isms in various projects for many years. And what I figured out is that they are recurring problem that fixing one single use of `[[` will never fix the next build failures. Those who use bash-specific syntax usually run /bin/bash on MacOS X. Unless you are using different shell like dash, you will never notice that. And you cannot ask them to use dash instead of bash.

Switching to explicit /bin/bash instead of /bin/sh is less disruptive to those who run dash as the default shell (I do). You could just install bash and have symlink at /bin/bash if you don't have it. I'm sure > 99% of Linux users could do it.

drowaudio:
This has been fixed on the develop branch for some time. We'll merge it in to master once the develop branch has been stable for a while. Thanks for pointing out the differences.

PR #15, opened 2020-02-11
Fix segfault when application is derived from JUCEApplicationBase

CharlesHolbrow:
The `updateJobs` method of the `tracktion_engine::BackgroundJobs` class segfaults if the application class derived from JUCEApplicationBase and not JUCEApplication.

The `JUCEApplication::getInstance()` is not virtual. However, it does just delegates to `JUCEApplicationBase::getInstance`.  See Juce's [`JUCEAapplication::getInstance()`](https://github.com/WeAreROLI/JUCE/blob/06c3674e510f991d88f881203aedd85a60fbced9/modules/juce_gui_basics/application/juce_Application.cpp#L34-L37)

This one line fix should allow any application derived from `JUCEApplicationBase` to use the BackgroundJobs class. 

See also: [forum.juce.com thread](https://forum.juce.com/t/segfault-when-application-is-derived-from-juceapplicationbase-and-not-juceapplication/37567?u=audioishi)

### DIFF
--- a/modules/tracktion_engine/model/edit/tracktion_EditUtilities.cpp
+++ b/modules/tracktion_engine/model/edit/tracktion_EditUtilities.cpp
@@ -335,6 +335,16 @@ void deleteRegionOfTracks (Edit& edit, EditTimeRange rangeToDelete, bool onlySel
         {
             t->deleteRegion (rangeToDelete, selectionManager);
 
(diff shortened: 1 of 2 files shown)


drowaudio:
Added now. Thanks.

PR #17, opened 2020-03-14
PlaybackDemo's playPauseButton won't change when it is clicked without command line parameter.

Dicklessgreat:
playPauseButton won't change its text when it is clicked without this line.


drowaudio:
Thanks, I've added this now. It's just clearing the CI.

PR #22, opened 2020-07-18
Doxygen: Fix make doc batch command.

COx2:
If don't set the value of the subdirs option, the pre-processing by python doesn't seem to work properly.


COx2:
Sorry, there were some commits differed issues mixed in.

PR #23, opened 2020-07-18
examples: Preventing some variables from becoming a NaN.

COx2:
For example, if the DAW's beat setting is 4/32nds, the quarterNotesPerBar variable will be 0, and the beats variable will be assigned as NaN. It causing a crash of VST host application.


drowaudio:
Thanks for the report and fix!

This has been added now here: https://github.com/Tracktion/tracktion_engine/commit/ed314bb23716cacec74e1c016c40f66570fc54ac

COx2:
Thanks.

PR #28, opened 2020-10-28
Fix all doxygen engine doc and also python script crashing because juce module header build/juce.h is not found

fab672000:
and gracefully indicate that the module header (i.e. the juce dependency) is skipped.

Also adds a convenient debug launch.json vscode python config file to allow debugging python script(s). In order to use vscode to build/debug python just install in vscode the "Python" and "Python for VSCode" extensions then open a python file.

In future, we should be able to generate the modules grouping directly from source and therefore not duplicate the full source code to the build folder, this seems unnecessary and slows down significantly the documentation process.
The usual approach for adding modules groups is to ask each class to add itself to a group (using the @addgroup statement) so that doxygen will automatically discover them eliminating the need for the script, while preserving complete genericity to that process.

For now keep using the script and focus of the documentation content.


examples/common/Utilities.h
```diff
@@ -15,8 +15,7 @@ namespace Helpers
 {
     static inline void addAndMakeVisible (Component& parent, const Array<Component*>& children)
     {
-        for (auto c : children)
-            parent.addAndMakeVisible (c);
+        for (auto* c : children) parent.addAndMakeVisible (c);
```
julianstorer, line 18: We don't write single-line for loops in this style (except where there are multiple loops on consecutive lines where lining them up vertically would be helpful to a reader).

Almost all the code in the repo is already formatted the way we like it, so if you see something like this, then maybe assume that we don't want it "fixed" unless you're 100% certain that it doesn't match the rest of the library.

Also, although in this case changing `auto` to `auto*` is fine, but there are many places in the code where `auto` is used in a for loop for possibly non-obvious but important reasons, so in general, steer clear of doing this unless it fixes a real problem!

fab672000: OK  I'll be more careful for the new line removal and respectful of whatever code style is in place.

Personally, I tend to avoid new lines without new brackets as I can't tell you how many bugs I found by people not careful of adding them. so I only allow myself to not add a bracket if the code is trivial and on the same line to avoid this:

if (my_enthusiastic_new_feature_works)
  TRACKTION_LOG("It should log this");
  TRACKTION_LOG("But not this unless I bug because it is indented but will always be called!");

Also for reading other's code, and for const and * alterations to auto: they help me a lot figuring what the code does and especially does not do (for const). This is part of the offical llvm recommendation BTW, see:
  https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-qualified-auto.html

fab672000: Applied the suggested change, thanks for clarifying about teh coding style, we should probably add a .dox page on this ;-)

modules/tracktion_engine/utilities/tracktion_Engine.h
```diff
@@ -33,15 +45,15 @@ class Engine
...
-    TemporaryFileManager& getTemporaryFileManager() const;
-    AudioFileFormatManager& getAudioFileFormatManager() const;
-    PropertyStorage& getPropertyStorage() const;
-    UIBehaviour& getUIBehaviour() const;
-    EngineBehaviour& getEngineBehaviour() const;
-    DeviceManager& getDeviceManager() const;
-    MidiProgramManager& getMidiProgramManager() const;
-    ExternalControllerManager& getExternalControllerManager() const;
-    RenderManager& getRenderManager() const;
+    TemporaryFileManager& getTemporaryFileManager() const; ///< Get the shared temporary file manager allowing to handle default app and user temporary folders.
```
julianstorer, line 48: Most of the comments in this section have grammatical mistakes that will jump out to any native English speakers as being wrong... 

I really hate to criticise something like that, and if it was just a comment on some internal code, I'd ignore it because it's clear what you meant.. but any new content that's intended for end-user-facing documentation really needs to be 100% correct.

modules/tracktion_engine/utilities/tracktion_Engine.h
```diff
@@ -33,15 +45,15 @@ class Engine
...
+    AudioFileFormatManager& getAudioFileFormatManager() const; ///< Get the audio files format manager that maintains a list of available wave file formats.
+    PropertyStorage& getPropertyStorage() const; ///< Get the top level User settings customizable xml file.
+    UIBehaviour& getUIBehaviour() const; ///< Get the engine UI elements customization helper class allowing to change default ui engine elements.
+    EngineBehaviour& getEngineBehaviour() const; ///< Get engine customization helper class allowing to change the engine default behavior.
+    DeviceManager& getDeviceManager() const; ///< Get the top level class for handling audio / midi devices @see juce::AudioDeviceManager.
+    MidiProgramManager& getMidiProgramManager() const; ///< Get the top level class that handles midi banks, programs, sets or presets.
```
julianstorer, line 54: MIDI should always be all-caps.

There's probably lots of lower-case uses of it in the library, but should get this right in all new content, and fix old ones when we see them (only in text, not necessarily in the code itself)

modules/tracktion_engine/utilities/tracktion_Engine.h
fab672000: OK, will sure revisit all use of MIDI and propose fixes ASAP, for the comment backward style any remarks?

i.e.  do you like the backward comment style (the idea is to not disrupt too much the code length and let those not interested to read focus on the left-side.

For grammar, will also be keen on any remark to make it better too, thanks for that.

modules/tracktion_engine/utilities/tracktion_Engine.h
fab672000: Applied suggested changes, thanks for helping me improve my (non native ) english too.

fab672000 (commented):
Also added all the doxygen warnings fixes in the engine and a new coding-style doxygen page from Jules original coding style article.

fab672000:
For legal concerns: I am aware and agree that any material improvement that I provide in this PR for the project (and also future ones if any here) can be freely used both non-commercially and commercially in accordance with the tracktion engine dual license.

drowaudio:
Added these here: https://github.com/Tracktion/tracktion_engine/commit/7acaf7933e5b8e607a67f2dcbe4f5ab943b3cb12

PR #32, opened 2020-12-10
Fixed compilation when enabling AirWindows with MSVC.

jrlanglois:
`M_PI` support needs to be explicitly enabled for MSVC (or so it seems!).

### DIFF
--- a/modules/tracktion_engine/tracktion_engine_airwindows.cpp
+++ b/modules/tracktion_engine/tracktion_engine_airwindows.cpp
@@ -9,6 +9,9 @@
 
 #if ! JUCE_PROJUCER_LIVE_BUILD
 
+#undef _USE_MATH_DEFINES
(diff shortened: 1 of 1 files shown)


modules/tracktion_engine/tracktion_engine_airwindows.cpp
```diff
@@ -50,7 +53,7 @@ using namespace juce;
 
 #if JUCE_WINDOWS
  #pragma warning (push)
- #pragma warning (disable : 4244 4100 4305 4065 4701 4706 4723)
+ #pragma warning (disable : 4244 4100 4305 4065 4701 4706 4723 4458)
```
jrlanglois, line 56: Also had shadowing warnings...

jrlanglois:
Please note that the error from your CI does not pertain to this PR:

image

PR #33, opened 2020-12-10
Fixed compilation when enabling control surfaces and disabling MackieC4 support.

jrlanglois:
And fixed shadowing warnings.


modules/tracktion_engine/control_surfaces/types/tracktion_AlphaTrack.cpp
```diff
@@ -696,7 +696,7 @@ void AlphaTrackControlSurface::handleAsyncUpdate()
     if (flip)
         updateFlip();
 
-    if (auto edit = getEdit())
+    if (auto e = getEdit())
```
jrlanglois, line 699: I'm a bit surprised that the `Edit*` member isn't in the `protected` section instead of `private`...

FigBug:
Fixed in 15a615964f7a4f23ac50ad265152f10fe7931e55

Issue #34, opened 2020-12-10
TRACKTION_ASSERT_MESSAGE_THREAD is redundant to JUCE_ASSERT_MESSAGE_MANAGER_IS_LOCKED


jrlanglois:
Obviously just a minor thing..

PR #36, opened 2020-12-10
Replace TRACKTION_ASSERT_MESSAGE_THREAD with the JUCE rendition.

jrlanglois:
See https://github.com/Tracktion/tracktion_engine/issues/34

This is simply a find/replace. JUCE's macro now offers a pointer check for the singleton instance as well.


FigBug:
These aren't equivalent. TRACKTION_ASSERT_MESSAGE_THREAD  is the same as JUCE_ASSERT_MESSAGE_MANAGER_IS_LOCKED

drowaudio:
I don't think these are identical as the Tracktion one will also return true if the current thread has a _locked_ message manager. I think the equivalent would be `JUCE_ASSERT_MESSAGE_MANAGER_IS_LOCKED`.

But... I'm not sure if `JUCE_ASSERT_MESSAGE_THREAD` could be used instead. I'm just weary of changing existing behaviour (there's all sorts of stuff that happens when loading an Edit on a background thread that this could interact with).

jrlanglois:
Good point about the macro not being equivalent - I _did_ miss the `*_IS_LOCKED` version...

PR #45, opened 2021-01-04
EditTimeRange: Added constexpr and noexcept where possible, and added a couple more methods.

jrlanglois:


julianstorer (commented):
With respect Joel, this sort of PR is just unnecessary noise where the effort for us to review it and give you feedback outweighs the value that it'd add.

The noexcept additions won't make any difference to performance, and you've even put some of them on functions that contain assertions, which could end up throwing if the client has implemented assertions using exceptions (I do that myself in some projects). Adding noexcept all over the juce codebase was a mistake which we're avoiding in this library.

Adding empty doxygen comments is also not helpful, especially where there are idiomatic C++ constructors involved, where it's obvious to a reader what their purpose is, and our style is to just keep them simple without any extra clutter.

Useful comments which add missing information for a reader: yes please. Empty or trivial ones: no thanks.

jrlanglois:
No worries! This is all reasonable sounding to me so I'll just close 'er up.

PR #52, opened 2021-04-10
Support JUCE latest version

vvruspat:


drowaudio:
If you use the tracktion_graph branch this should be fixed already.
We'll be merging that in to master once all the minor issues have been resolved. (You can still use the old engine in this branch).

vvruspat:
> If you use the tracktion_graph branch this should be fixed already.
> We'll be merging that in to master once all the minor issues have been resolved. (You can still use the old engine in this branch).

Cool!

PR #64, opened 2021-09-09
Feature/plugin demo

drowaudio:


tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
+# 04 - Distortion Effect
+
+In this tutorial, we're going to analyze the structure of an audio effect application, by applying distortion to a looping audio file.
+
+## Main Tracktion Engine Classes
+#### Engine
```
drowaudio, line 6: As this is part 4 of a series. You probably don't need to repeat the classes that have already been covered.
What should probably be here though are `Plugin` and `PluginManager` as I don't think they're in previous tutorials.

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### Track
+'Edit' contains a list of 'Track's. This project makes use of the 'AudioTrack' which outputs audio.
+
+#### Clip
+'Clip' is the base class for clips that live in 'Track's. 'Clip's can be MIDI, audio, step or the edit-in-an-edit 'editClip.'
+
+
+## Main JUCE Classes
+#### Plugin
```
drowaudio, line 20: `Plugin` isn't a JUCE class so this section can probably be removed.

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
+# 04 - Distortion Effect
+
+In this tutorial, we're going to analyze the structure of an audio effect application, by applying distortion to a looping audio file.
```
drowaudio, line 3: Also probably state what key topics covered are here such as:
- Adding customisation to an Edit through use of a custom Plugin class
- Writing this plugin class
- Registering it with the Engine so it can be added to an Edit's PluginList
- Adding it to a track
- Writing a custom ValueSource to bind a Slider/Value to an AutomatableParameter

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+'Clip' is the base class for clips that live in 'Track's. 'Clip's can be MIDI, audio, step or the edit-in-an-edit 'editClip.'
+
+
+## Main JUCE Classes
+#### Plugin
+
+## DistortionEffectDemo.h
+
+
+## DistortionPlugin Class
```
drowaudio, line 25: I'd add an into as to what the DistortionPlugin class is here. I.e. a Plugin subclass so it can be added as a plugin to clips/tracks etc. in an Edit. For demo purposes we're using just a very simple tanh waveshaper

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+
+### DistortionPlugin() Constructor
+
+- In the constructor, we initialise the undo manager to enable users to undo plugin value changes, and refer each plugin value to the undo manager.
```
drowaudio, line 29: This isn't really correct. We're not initialising the UndoManager.
We're referring the gainValue CachedValue member to the plugin's `state` gain` property, using the Edit's undo manager and providing a default value of 1.

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+```
+auto um = getUndoManager();
+gainValue.referTo (state, IDs::gain, um, 1.0f);
+```
+
+- We then initialise each plugin parameter and attach them to the values they represent.
```
drowaudio, line 35: Probably better to more accurate and rather than "to the values they represent" say something like "We create an AutomatableParameter, assign it to our gainParam member. `addParam` registers the parameter with the `Plugin`".
Then something like the `attachToCurrentValue` call attaches the `gainValue` `CachedValue` member to the gain parameter.

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+
+- The `applyToBuffer` function is where plugin processing takes place. The destination buffer can be obtained from the `PluginRenderContext` in order to apply digital signal processing.
+
+- First, if the plugin is disabled, we return the function early to prevent processing.
+```
+if (! isEnabled())
+    return;
+```
+
+- We iterate through each sample of each channel and apply the tanh waveshaping function to the output audio.
```
drowaudio, line 59: Maybe also add something to show we're using the gain parameter's current value to boost the signal before shaping?

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+    auto dest = fc.destBuffer->getWritePointer (channel);
+                
+    for (int i = 0; i < fc.bufferNumSamples; ++i)
+        dest[i] = std::tanh (gainValue * dest[i]);
+}
+```
+
+## DistortionEffectDemo Class
+
+The DistortionEffectDemo is defined similarly to the PitchAndTimeComponent class
```
drowaudio, line 72: Maybe extend this slightly to state that this demo class just loads an example file and loops it, applying the above DistortionPlugin to the track's output?

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+```
+auto gainPlugin = edit.getPluginCache().createNewPlugin (DistortionPlugin::xmlTypeName, {});
+track->pluginList.insertPlugin (gainPlugin, 0, nullptr);
+```
+- Finally, in order to hear the effect being played, we must set the transport loop range to the start/end of the clip, set it to loop, and initialise play.
+```
+edit.getTransport().addChangeListener (this);
+EngineHelpers::loopAroundClip (*clip);
+```
+
```
drowaudio, line 133: Before you write this I'd add a section on the `ParameterValueSource` class explaining what it does and why it needs to listen to the AutomatableParameter. You can use the two sliders as an example so they stay in sync. In a real DAW you can have multiple representations of a parameter e.g. a curve (which will just be straight line if it has no points), a rotary slider in the properties panel parameter list, a quick control slider on the plugin mixer UI element and so on. Basically this is a quick, concise way of attaching Value backed Components to AutomatableParameters.

tutorials/04 - DistortionEffectDemo.md
```diff
@@ -0,0 +1,138 @@
...
+- In order to control the gain parameter with the slider, we need to bind them together.
+```
+auto gainParam = gainPlugin->getAutomatableParameterByID ("gain");
+bindSliderToParameter (gainSlider, *gainParam);
```
drowaudio, line 137: Also elaborate on this a bit. What does "bind them together mean"?

drowaudio (commented):
Generally the structure here is good, it just needs fleshing out in places to describe the new concepts introduced here.

PR #66, opened 2021-10-01
RubberBand:   Added initial RubberBand classes for dev

drowaudio:


modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+  '-.  .-'|  .--' ,-.  | .--'|     /'-.  .-',--.| .-. ||      \   Tracktion Software
+    |  |  |  |  \ '-'  \ `--.|  \  \  |  |  |  |' '-' '|  ||  |       Corporation
+    `---' `--'   `--`--'`---'`--'`--' `---' `--' `---' `--''--'    www.tracktion.com
+
+    Tracktion Engine uses a GPL/commercial licence - see LICENCE.md for details.
+*/
+
+#include "tracktion_RubberBand.h"
+
+#ifdef __GNUC__
```
drowaudio, line 13: These were required to silence warnings on clang on macOS but they might need to be refined for GCC on Linux and will need a different set on Windows

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+ #pragma GCC diagnostic ignored "-Wunused-variable"
+ #pragma GCC diagnostic ignored "-Wunused-parameter"
+ #pragma GCC diagnostic ignored "-Wpedantic"
+ #pragma GCC diagnostic ignored "-Wunknown-pragmas"
+#endif
+
+#define Point CarbonDummyPointName
+#define Component CarbonDummyCompName
+
+//#include "rubberband/rubberband-c.cpp"
```
drowaudio, line 34: If these commented out files aren't needed, the lines can be removed

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+#include "../3rd_party/rubberband/src/RubberBandStretcher.cpp"
+
+#undef Point
+#undef Component
+
+#ifdef __GNUC__
+ #pragma GCC diagnostic pop
+#endif
+
+bool RubberStretcher::setSpeedAndPitch(float speedRatio, float semitonesUp)
```
drowaudio, line 67: There should be spaces before an opening parenthesis e.g.:
`setSpeedAndPitch (float speedRatio`
`float pitch = juce::jlimit (0.25f`

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+{
+    float pitch = juce::jlimit(0.25f, 4.0f, tracktion_engine::Pitch::semitonesToRatio (semitonesUp));
+
+    // Not sure what to do with sync
+
+    // bool sync = (elastiqueMode == TimeStretcher::elastiquePro) ? elastiqueProOptions.syncTimeStrPitchShft : false;
+
+    // TODO: Assert this works
+    thisRubberBandStretcher.setPitchScale(pitch);
```
drowaudio, line 76: `thisRubberBandStretcher` isn't a great name. What does "this" mean? It should probably just be called `rubberBandStretcher`

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+    thisRubberBandStretcher.setTimeRatio(speedRatio);
+
+    int r = (thisRubberBandStretcher.getPitchScale() == pitch && thisRubberBandStretcher.getTimeRatio() == speedRatio);
```
drowaudio, line 79: The `int r =` type is wrong here. Both of the following are boolean expressions so `r` should be a `const bool r = `

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+    jassert(r == 0); juce::ignoreUnused(r);
```
drowaudio, line 80: More spaces here and with the above change it should be:
`jassert (r); juce::ignoreUnused (r);`

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+}
+
+int RubberStretcher::getMaxFramesNeeded() const
+{
+    return maxFramesNeeded;
+}
+
+void RubberStretcher::processData(const float* const* inChannels, int numSamples, float* const* outChannels)
+{
+    thisRubberBandStretcher.process((float**)outChannels, numSamples, false);
```
drowaudio, line 99: This doesn't look right to me. The docs say `Provide a block of "samples" sample frames for processing`, outChannels is where the output data should be, i.e. the time-stretched/pitched data.

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+    TempBuffer.setDataToReferTo((float**)inChannels, 2, numSamples);
```
drowaudio, line 101: The case of `TempBuffer` is wrong here, it should start with a lower case letter as it's not a type.
But I don't actually think a member is required? Why isn't the `inChannels` just passed here?

Also, the loop looks wrong, you're checking how many samples are available to retrieve from rubber band but then just processing the same samples in to the stretcher. That doesn't make any sense.
It seems like you should be doing this check to see if you can actually do the `retrieve` step?

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+        runPitchShiftTest();
+        runTimeStretchTest();
+    }
+
+private:
+    void runPitchShiftTest()
+    {
+        beginTest ("Pitch shift");
+        
+        RubberStretcher thisRubberBandStretcher = RubberStretcher (44100, 512, 2, RubberBandStretcher::Option::OptionProcessRealTime);
```
drowaudio, line 265: This declaration has some redundant repetition. It can simply be written as `RubberStretcher thisRubberBandStretcher (44100, 512, 2, RubberBandStretcher::Option::OptionProcessRealTime);`

It's also probably better to use a constant variable for the sample rate, buffer size and num channels are they're used several times in the method.

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+        thisRubberBandStretcher.reset();
+
+        AudioBuffer<float> sinBuffer;
+        sinBuffer.setSize(2, 44100);
+
+        auto wL = sinBuffer.getWritePointer(0);
+        auto wR = sinBuffer.getWritePointer(1);
+
+        double currentAngle = 0.0, angleDelta = 0.0;
+        float originalPitch = 432.0f; //A4
```
drowaudio, line 278: Isn't `440.0f` typically used for A at concert pitch?

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,365 @@
...
+        for (int sample = 0; sample < 44100; ++sample)
+        {
+            auto currentSample = (float)std::sin(currentAngle);
+            currentAngle += angleDelta;
+            wL[sample] = currentSample;
+            wR[sample] = currentSample;
+        }
+
+        // Check number of zero crossings before - This line might not even be necessary. Just holding onto data
+        // int numZeroCrossings = getNumZeroCrossings (sinBuffer);
```
drowaudio, line 292: Remember commented out code shouldn't be committed

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
```diff
@@ -0,0 +1,112 @@
...
+{
+    RubberStretcher(double sourceSampleRate, int samplesPerBlock, int numChannels, RubberBand::RubberBandStretcher::Option option)
+                  : thisRubberBandStretcher(sourceSampleRate, numChannels, option, 1.0f, 0.5f)
+
+    {
+
+        sampleRate = sourceSampleRate;
+        TempBuffer.setSize(numChannels, sourceSampleRate);
+
+        if (&thisRubberBandStretcher == nullptr)
```
drowaudio, line 47: This is a redundant check as a pointer to a member object will never be nullptr. You only need to check _pointers_ if they are `nullptr`.

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
```diff
@@ -0,0 +1,112 @@
...
+        {
+            jassertfalse;
+            TRACKTION_LOG_ERROR("Cannot create Rubber Band Stretcher");
+        }
+        else
+        {
+            // This must be called first, before any other functions and can not
+            // be called again
+            maxFramesNeeded = thisRubberBandStretcher.available();
```
drowaudio, line 56: This doesn't seem to make any sense. If you've just created the stretcher, it will always have 0 frames available and hence store that to `maxFramesNeeded`? `maxFramesNeeded` is used to determine how much latency we need to introduce for a real-time pitch shift which I'm sure won't be zero. That would indicate that for any number of input samples, the same number of output samples will be generated.

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        sampleRate = sourceSampleRate;
+
+        thisRubberBandStretcher.reset();
+    }
+
+    ~RubberStretcher() override
```
drowaudio, line 64: An empty destructor probably isn't required, especially in the class declaration as the compiler will just generate one.

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
```diff
@@ -0,0 +1,112 @@
...
+    {
+
+    }
+
+    bool isOk() const override { return &thisRubberBandStretcher != nullptr; }
```
drowaudio, line 69: Again this is a redundant check (there's even compiler warnings about it). Is there another rubber band method that can be used to check that it's initialised correctly?

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
```diff
@@ -0,0 +1,112 @@
...
+
+    void reset() override { thisRubberBandStretcher.reset(); }
+
+    bool setSpeedAndPitch(float speedRatio, float semitonesUp) override;
+    int getFramesNeeded() const override;
+    int getMaxFramesNeeded() const override;
+
+    void processData(const float* const* inChannels, int numSamples, float* const* outChannels) override;
+
+    bool testPitchShift(float tolerance);
```
drowaudio, line 79: I've removed these unused methods in the next commit and moved the ones used for testing in to the actual test as they don't require any state from this object.

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
```diff
@@ -0,0 +1,112 @@
...
+    int getSamplesRequired()
+    {
+        return thisRubberBandStretcher.getSamplesRequired();
+    }
+
+    void flush(float* const* outChannels) override
+    {
+    }
+
+    RubberBand::RubberBandStretcher thisRubberBandStretcher;
```
drowaudio, line 99: As I mentioned before this is an odd name. It should probably just be `RubberBand::RubberBandStretcher rubberBandStretcher;`.

It should also probably be in the private section as no other classes should need to have access to it?

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
```diff
@@ -0,0 +1,112 @@
...
+
+private:
```
drowaudio, line 101: The spacing here is a bit odd and as I mentioned before I don't think TempBuffer is needed.

drowaudio (commented):
There's quite a bit to look over here and I'm not convinced it actually works as expected. The process method seems like it's in the wrong order.

Additionally there are lots of warnings in these files, these need cleaning up. Don't worry about warnings in 3rd party code, we silence those or we'll have to edit the code every time we update the library.

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        // Check whether buffer has resized
+        expectEquals (scount, 22050);
+        expectLessThan (std::abs (sinBuffer.getNumSamples() - testBuffer.getNumSamples() * 2), 10);
+    }
+
+    //==============================================================================
+    float getPitchFromNumZeroCrossings (int numZeroCrossings, int numSamples, double sampleRate)
+    {
+        float bufferTimeInSeconds = numSamples / sampleRate;
```
drowaudio, line 240: The spacing doesn't look right here and these should probably be `const` if they're not being mutated.
There is also automated type conversion (dividing an int by a double and then assigning it to a float), generally we want all type conversions to be explicit. There should be warnings about this.

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        int numCycles = numZeroCrossings / 2;
```
drowaudio, line 242: Should this really be an int? What if there are an odd number of zero crossings?

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        float pitchInHertz = numCycles / bufferTimeInSeconds;
+
+        return pitchInHertz;
+    }
+
+    int getNumZeroCrossings (AudioBuffer<float> buffer)
```
drowaudio, line 249: This `AudioBuffer<float> buffer` is taking a copy of the buffer which is unnecessarily expensive. Because this buffer isn't mutated in any way by the function it should be a const reference.

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -0,0 +1,269 @@
...
+        auto dest = buffer.getWritePointer(0);
+
+        int numZeroCrossings = 0;
+
+        for (int sample = 1; sample < buffer.getNumSamples(); ++sample)
+        {
+            if ((dest[sample] < 0.0f && dest[sample - 1] > 0.0f)
+                || (dest[sample] > 0.0f && dest[sample - 1] < 0.0f))
+            {
+                numZeroCrossings++;
```
drowaudio, line 260: You can use pre-increment here as you don't need the value before the increment.

drowaudio (commented):
A few more notes on the code I moved around.

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -207,46 +199,44 @@ class RubberBandStretcherTests  : public UnitTest
...
-                scount++;
+                ++zeroCountR;
             }
         }
 
         // Check whether buffer has resized
-        expectEquals (scount, 22050);
-        expectLessThan (std::abs (sinBuffer.getNumSamples() - testBuffer.getNumSamples() * 2), 10);
+        expectEquals (zeroCountL, 22050);
+        expectEquals (zeroCountR, 22050);
```
drowaudio, line 224: This is a good check to do but it doesn't check if the pitch is the same as the original. This is what we need to check for a time-stretch (a pitch and time change would be effectively a resample).

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
```diff
@@ -207,46 +199,44 @@ class RubberBandStretcherTests  : public UnitTest
...
-        int numCycles = numZeroCrossings / 2;
+        float numCycles = numZeroCrossings / 2.0f;
 
         float pitchInHertz = numCycles / bufferTimeInSeconds;
 
         return pitchInHertz;
     }
 
-    int getNumZeroCrossings (AudioBuffer<float> buffer)
+    int getNumZeroCrossings (AudioBuffer<float>& buffer)
```
drowaudio, line 239: This is better but it should be a `(const AudioBuffer<float>& buffer)` to show that we're not changing `buffer`

drowaudio (commented):
I've added some more comments. You might want to go back to the original ones I left as it seems like some have dropped off the review as the lines have changed.

modules/tracktion_engine/timestretch/tracktion_RubberBand.h
godofecht: I'm not completely sure about this. I've looked through the method list but RubberBand doesn't have any obvious intialization methods. There is a variable called _initalize but it is inaccessible and I'm not sure what it does.

What I'm currently thinking is checking to see if rubberBandStretcher.available() == 0

modules/tracktion_engine/timestretch/tracktion_RubberBand.cpp
godofecht: Got it. So we would want to calculate the pitch of the converted buffer and compare it to the pitch of the original buffer by comparing the number of zero crossings.

PR #80, opened 2021-11-01
Apply edge fades

drowaudio:


modules/tracktion_engine/utilities/tracktion_EngineBehaviour.h
```diff
@@ -178,6 +178,8 @@ class EngineBehaviour
         1: loopRangeDefinesSubsequentRepetitions    // The first section is the whole sequence, subsequent repitions are determined by the loop range.
     */
     virtual int getDefaultLoopedSequenceType()                                      { return 0; }
+
+    virtual bool autoAddClipEdgeFades()                                             { return true; }
```
drowaudio, line 182: This needs a doxygen comment describing what the function does. See the other methods here for an example of the detail level required. Basically just describe what returning true or false does here.

This also needs to be changed back to `false` by default for merging back in to develop to not change the behaviour for existing users.

drowaudio (commented):
You've also managed to change the JUCE submodule commit here. You need to make sure you don't do that when creating branches or it will make merging back difficult.

PR #91, opened 2021-12-07
Midi bug:   Added discardRecordings method to Consumer

drowaudio:


modules/tracktion_engine/playback/devices/tracktion_InputDevice.h
```diff
@@ -262,6 +262,8 @@ class InputDeviceInstance   : protected juce::ValueTree::Listener
 
         /** Override this to receive MIDI input from the device if it has any. */
         virtual void handleIncomingMidiMessage (const juce::MidiMessage&) {}
+
+        virtual void discardRecordings() {}
```
drowaudio, line 266: This needs a comment describing what the method should do (see the other virtual methods.
Explain this is called when a recording is aborted so listeners should trash their temp data.

modules/tracktion_engine/playback/devices/tracktion_MidiInputDevice.cpp
```diff
@@ -717,7 +717,11 @@ class MidiInputDeviceInstanceBase  : public InputDeviceInstance
         CRASH_TRACER
 
         if (discardRecordings)
+        { 
+            for (auto c : consumers)
+                c->discardRecordings();
             recorded.clear();
```
drowaudio, line 723: Leave a space above `recorded.clear()`

modules/tracktion_engine/playback/graph/tracktion_MidiInputDeviceNode.cpp
```diff
@@ -202,4 +202,9 @@ bool MidiInputDeviceNode::isLivePlayOverActive()
     return instance.isRecording() && instance.context.transport.looping;
 }
 
+void MidiInputDeviceNode::discardRecordings()
+{
+    liveRecordedMessages.clear();
```
drowaudio, line 207: It won't be thread safe to call this as the audio thread could still be writing to the array.
Guard it with a `const juce::ScopedLock sl (liveInputLock);` like the other uses in the class.

modules/tracktion_engine/playback/graph/tracktion_MidiInputDeviceNode.h
```diff
@@ -49,7 +50,7 @@ class MidiInputDeviceNode final : public tracktion_graph::Node,
     //==============================================================================
     void processSection (ProcessContext&, juce::Range<int64_t> timelineRange);
     void createProgramChanges (MidiMessageArray&);
-    bool isLivePlayOverActive();
+    bool isLivePlayOverActive();
```
drowaudio, line 53: I think you've added an accidental space on the end here.

drowaudio (commented):
Some minor tweaks to be made here.

Issue #96, opened 2021-12-17
Issues when trying to build examples


cannoneyed:
When generating the example projects on MacOS 10.15.4(Xcode 11.5), the `TestRunner` demo fails with the following error
```
Building example: TestRunner
Re-saving file: /Users/andycoenen/cannoneyed/tracktion_engine/examples/projects/TestRunner/TestRunner.jucer
Error when saving: At least one of your modules has missing dependencies!
Please add the required dependencies, or run the command again with the "--fix-missing-dependencies" option.
```

This causes the build script to stop early. Opening up the .jucer file in the ProJucer, the error is being caused by the `tracktion_engine` module requiring the `tracktion_graph` module to be added. The issue can be fixed by simply adding `tracktion_graph` as a dependency to each of the PIPs. 

https://github.com/Tracktion/tracktion_engine/pull/97

Also, the generated projects don't compile without setting the preprocessor flag `JUCE_MODAL_LOOPS_PERMITTED=1`. This can be done quickly in the ProJucer but isn't being set via the PIP

drowaudio:
Yeah I know this is a problem. It's fixed on the develop branch though. I've got a couple more tweaks to do and then I'll merge develop in to main. In the mean time I'd stick with the develop branch.

PR #97, opened 2021-12-17
Add tracktion_graph as a dependency to the example PIPs

cannoneyed:
When generating the example projects on MacOS 10.15.4(Xcode 11.5), the `TestRunner` demo fails with the following error
```
Building example: TestRunner
Re-saving file: /Users/andycoenen/cannoneyed/tracktion_engine/examples/projects/TestRunner/TestRunner.jucer
Error when saving: At least one of your modules has missing dependencies!
Please add the required dependencies, or run the command again with the "--fix-missing-dependencies" option.
```

This causes the build script to stop early. Opening up the .jucer file in the ProJucer, the error is being caused by the `tracktion_engine` module requiring the `tracktion_graph` module to be added. The issue can be fixed by simply adding `tracktion_graph` as a dependency to each of the PIPs. 

See https://github.com/Tracktion/tracktion_engine/issues/96


drowaudio:
This is fixed on develop now.

PR #99, opened 2021-12-27
Add a missing null check for projectItem when processing recordings.

jklein24:
This check is a couple places in this function already, but this one was missing.


drowaudio:
Cheers, this should be fixed here: https://github.com/Tracktion/tracktion_engine/commit/6ccabdcebc4a437cfd88a32fa628cf1445036eaa

PR #103, opened 2022-01-07
RackType::addPlugin returns a boolean indicating if the plugin was added

adamnemecek:


drowaudio:
Added here: https://github.com/Tracktion/tracktion_engine/commit/fa5eb48c43f61b6539f9057e95cc0d8bbca170bf

PR #110, opened 2022-01-27
Time:   Added some time primitives

drowaudio:


modules/tracktion_engine/model/automation/tracktion_AutomationCurve.cpp
```diff
@@ -158,7 +161,7 @@ float AutomationCurve::getValueAt (double time) const
     if (curve1 >= -0.5f && curve1 <= 0.5f)
     {
         auto bezierPoint = getBezierPoint (index - 1);
-        return getBezierYFromX (time, time1, value1, bezierPoint.time, bezierPoint.value, time2, value2);
+        return getBezierYFromX (time, time1, value1, toTime (bezierPoint.time, getOwnerParameter()->getEdit().tempoSequence).inSeconds(), bezierPoint.value, time2, value2);
```
julianstorer, line 164: worries me to see a pointer being dereferenced like that without checking it. Maybe getOwnerParameter() should return a reference, or there should be a method to return a reference to the tempoSequence?

modules/tracktion_engine/model/automation/tracktion_AutomationCurve.cpp
```diff
@@ -772,24 +778,24 @@ CurvePoint AutomationCurve::getBezierHandle (int index) const noexcept
...
     float y1end, y2end;
     getBezierEnds (index, x1end, y1end, x2end, y2end);
-    return { x1end, y1end };
+    return { TimePosition::fromSeconds (x1end), y1end };
 }
 
 void AutomationCurve::getBezierEnds (int index, double& x1out, float& y1out, double& x2out, float& y2out) const noexcept
 {
-    auto x1 = getPointTime (index);
+    auto x1 = getPointTime (index).inSeconds();
```
julianstorer, line 794: would it be better to make the function return position objects instead of doubles?

modules/tracktion_engine/model/automation/tracktion_AutomationCurve.cpp
drowaudio: Yeah, I knew this was a bit iffy when I wrote it. I was trying to think of a situation where an `AutomationCurve` wouldn't have a parameter. I couldn't think of any but it looks like there's a reason why `AutomationCurve::setOwnerParameter` is a separate function. Can you remember why this is?

Looking at the Waveform code, I con only see places where the parameter is set immediately after an `AutomationCurve` creation and I can't see anywhere where it's set to `nullptr` so it should probably just be passed in the constructor and returned as a reference.

modules/tracktion_engine/model/automation/tracktion_AutomationCurve.cpp
drowaudio: Yes... I do remember thinking that at the time but decided against it. But I can't remember why now. Probably because when I started I was trying to get it to build with minimal changes and then later abandoned that idea as it was more tricky to go half in. I'll update that.

modules/tracktion_engine/model/automation/tracktion_AutomationCurve.cpp
drowaudio: This is more complex than I thought. There's an assertion to ensure the parameter is valid before these calls are made though so I think it's safe to assume it will be later on in the function.

Issue #116, opened 2022-04-21
EqualizerPlugin.cpp lines 71->73 and 83->85 say 'High/Low-pass' when they should say 'High/Low-shelf'


jamierpond:
Hey guys, I think there are some typos in the Equalizer plugin cpp file. I think they are minor ones, but it changes the meaning of the label quite a lot!

Currently:
```
addAutomatableParameter (hiFreq = new EQAutomatableParameter (*this, "High-pass freq", TRANS("High-shelf freq"), *this, { 
```

Should be:
```
addAutomatableParameter (hiFreq = new EQAutomatableParameter (*this, "High-**shelf** freq", TRANS("High-shelf freq"), *this, { 
```

drowaudio:
Yeah maybe that would be a better name but it can't be changed now as that's the property the model references and is stored in the XML file. If we changed it, all existing sessions would break and it's not significant enough a change to add the the edit conversion pass (which would break minor version backwards compatibility which we rely on for testing).

jamierpond:
Oh damn, I see! Sorry I didn't see this before the PR. That makes sense. Would be cool to find a solution for this at some point!

PR #117, opened 2022-04-21
Fix incorrect label in tracktion_Equaliser.cpp

jamierpond:
This is corresponding to the issue I raised earlier. Please let me know if I've misunderstood the code! :)


drowaudio:
Thanks, but did you see the message I replied with? https://github.com/Tracktion/tracktion_engine/issues/116
It's a bit hard to change those tag names now I'm afraid.

PR #129, opened 2022-09-26
Pause if updating prepared node

jamierpond:
There seems to be some checks for this at the top of this function, but this doesn't really seem to be all that robust, as we still get crashes there, this just adds in more checks. I haven't been able to hit any crashes with these changes in conjuction with the other ScopedReallocationInhibitor.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,9 +20,6 @@ jobs:
           - name: Windows
             os: windows-latest
             platform: windows
-          - name: Windows_2016
-            os: windows-2016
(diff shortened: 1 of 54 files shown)


drowaudio:
I'm not sure what the status of this PR is but I think the original issue might be fixed by some refactoring I was doing in this PR: https://github.com/Tracktion/tracktion_engine/pull/127

Most notably I've introduced a `LockFreeObject` class in to `LockFreeMultiThreadedNodePlayer` which should avoid these issues and clean the code up a bit.

PR #130, opened 2022-10-07
Add TimePosition double constuctor in seconds

jamierpond:
Basically allows constructions of TimeRanges with branced loops, and implicit casting of doubles to TimePositions.


drowaudio:
I know it seems annoying but I deliberately didn't want to be able to construct a TimePosition from a double.
The whole point is that a TimePosition isn't in "seconds" so constructing one from a double is ambiguous. Is the double minutes, hours, frames etc. That's the whole point of the explicit `fromSeconds` function.

If you're using compile-time constants, you can use the chrono literals to construct them e.g. `1s, 2ms` etc.

drowaudio:
And can't you construct a BarsAndBeats using aggregate initialisation like this:
`BarsAndBeats ({ numBars })`?

Issue #105, opened 2022-01-14
AUv3 & ExternalPlugin on iOS


mathieugarcia:
Hi!

AUv3 plugins, on iOS, need to be instantiated asynchronically.

Right now, `ExternalPlugin::doFullInitialisation()` will end up calling JUCE's `PluginFormatManager::createPluginInstance()`, which will immediately return with the following error: _"This plug-in cannot be instantiated synchronously"._

It is possible to assign a specific lambda for custom instantiation on the global `PluginManager` (via the `createPluginInstance` variable), however, since this call is initially encapsulated in a `callBlocking` lambda, which internally relies on a `juce::WaitableEvent`, and since `PluginFormatManager::createPluginInstanceAsync` also uses a `juce::WaitableEvent`, the code is never triggered because of these competing blocking mechanisms. 

After some initial research, it seems that the best approach would be to patch `ExternalPlugin` instantiation/initialization code directly so it can create instances asynchronically and resume once it's ready. 

Before diving in, I'm just checking in with the team to see if it's maybe already in the backlog or, alternatively, to gather implementation ideas before submitting a pull request.

Thanks in advance,

Mathieu.

mathieugarcia:
_Update_: well, that was quick. I am delaying the initialization after the instance is returned, and this works perfectly. Still happy to hear your ideas so I can improve the code before submitting it.

drowaudio:
Can you show me what changes you need to make? Presumably if a `juce::AudioPluginFormatManager::createPluginInstance` returns a nullptr instance, you just asynchronously call `te::ExternalPlugin::doFullInitialisation`?

Tbh I've not had time to think about this in much detail.

mathieugarcia:
There's a bit more involved as `doFullInitialisation()` is initially called from the ctor (via `initialiseFully()`), so I added a simple mechanism to postpone and resume the initialisation once the instance is returned, in the lambda passed by `juce::AudioPluginFormatManager::createPluginInstanceAsync`. I also manually call `ExternalPlugin::initialise()` as the instance could be returned after the engine initially calls this method.

Here's the patch as of now - more testing is needed but so far it works nicely:

```
diff --git a/modules/tracktion_engine/plugins/external/tracktion_ExternalPlugin.cpp b/modules/tracktion_engine/plugins/external/tracktion_ExternalPlugin.cpp
index 4df1c907..44fb7bee 100644
[...]

PR #144, opened 2023-02-25
Added support for asynchronous instantiation of ExternalPlugins.

mathieugarcia:
Support for asynchronous plugin instantiation in `ExternalPlugin`. This allow proper hosting of AUv3 plugins on iOS and possibly other platforms and/or plugin formats that require asynchronous instantiation.

What changed:
- Added a static helper method `ExternalPlugin::requiresAsyncInstantiation()`.
- Moved post-initialisation code in a method called `completePluginInitialisation()`, to avoid code redundancy and minimize differences between the sync and async versions. This method is called once the plugin instance is available.
- `createPluginInstance` will check if the plugin requires asynchronous instantiation, if so, JUCE's `PluginFormatManager::createPluginInstanceAsync()` is called, with a callback that will perform all the necessary tasks after the plugin instance has been returned. Note that `baseClassInitialise()` and `initialise()` have to be called manually so the `ExternalPlugin` state is fully initialised.
- `createPluginInstance` method signature will not return the error String anymore.

For more information and discussion about this pull request see https://github.com/Tracktion/tracktion_engine/issues/105


modules/tracktion_engine/plugins/external/tracktion_ExternalPlugin.cpp
```diff
@@ -1707,22 +1726,38 @@ bool ExternalPlugin::setBusLayout (juce::AudioChannelSet set, bool isInput, int
...
-
-    juce::String error;
-    pluginInstance = engine.getPluginManager().createPluginInstance (description, dm.getSampleRate(), dm.getBlockSize(), error);
-
-    if (pluginInstance != nullptr)
+    
+    if (requiresAsyncInstantiation (engine, description))
+    {
+        // While instantiation is pending, make sure the plugin is disabled
+        setEnabled (false);
```
drowaudio, line 1738: I'm not sure setting the plugin as "disabled" is the right approach here, that basically means the same thing as bypassed and there's nothing to stop users un-bypassing plugins during the creation time period. It would also show in the UI that these plugins were "bypassed" when they're really just not ready for playback yet.
I think we might need a new state member "isCreatingInstance" to signify this which can be checked at the start of the process block (if that's what the setEnabled is for here?).

modules/tracktion_engine/playback/graph/tracktion_MidiInputDeviceNode.cpp
```diff
@@ -138,7 +138,7 @@ void MidiInputDeviceNode::processSection (ProcessContext& pc, juce::Range<int64_
     if (timeNow > lastReadTime + maxExpectedMsPerBuffer)
     {
         //jassertfalse
-        numMessages = 0;
+        //numMessages = 0;
```
drowaudio, line 141: I'm not sure why this is in this PR? Can it be removed so we can consider it separately?

modules/tracktion_engine/plugins/external/tracktion_ExternalPlugin.h
```diff
@@ -37,6 +37,8 @@ class ExternalPlugin  : public Plugin
     void forceFullReinitialise();
 
     juce::String getLoadError();
+
```
drowaudio, line 40: Does this need to be public? Perhaps it should be private?

modules/tracktion_engine/plugins/external/tracktion_ExternalPlugin.cpp
```diff
@@ -1707,22 +1726,38 @@ bool ExternalPlugin::setBusLayout (juce::AudioChannelSet set, bool isInput, int
...
+        engine.getPluginManager().pluginFormatManager
+          .createPluginInstanceAsync (description, dm.getSampleRate(), dm.getBlockSize(),
+                                      [&] (auto _instance, const auto& _err) -> void {
```
drowaudio, line 1741: Just a small style thing, we don't use leading underscores for name.
As the standard can reserve anything starting with an underscore in the global namespace (include `#define` etc) these can be problematic. Especially if they have generic names like `_err`. I'd just remove the leading underscores here.

drowaudio, line 1741: I don't think this can take a reference as the capture type. If the plugin gets deleted before the callback happens this will wreak havoc. Best to take a weak reference instead.
You can do this with `Selectable::getWeakRef()` and dynamic casting it but I think I'm going to add a wrapper class to do this for you automatically (like juce::Component::SafePointer) very soon so you might want to wait for that.

drowaudio (changes requested):
Great stuff. This is getting very close now.
I've just left a few bits of feedback that need addressing.

drowaudio:
FYI here's that Selectable helper class I mentioned: https://github.com/Tracktion/tracktion_engine/commit/c9b328c89f807dfa356778442a98bd6d8ce354c0

drowaudio:
Sorry, it's been a while since I looked at this and have forgotten most of the details. Plugin hosting with all the external formats is precarious at best which is why I'm hesitant to make changes without fully understanding all the ramifications.

I think the main outstanding issue was that you had to call:
```
    safeThis->baseClassInitialise (info);
    safeThis->initialise (info);
```
I couldn't understand why the audio graph isn't rebuilt automatically and then the `PluginNode` handle calling ` baseClassInitialise`? Any update on that front?

PR #155, opened 2023-05-10
MidiNode: Fixed a couple of loops and assertions regarding the range channelNumbers end

svenemtell:
See discussion here:
[[MidiNode channelNumbers range?](https://forum.juce.com/t/midinode-channelnumbers-range/56114)]

### DIFF
--- a/modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
+++ b/modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
@@ -552,8 +552,7 @@ namespace MidiHelpers
 
             controllerMessagesScratchBuffer.clearQuick();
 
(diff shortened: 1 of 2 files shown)


drowaudio:
Thanks. I think it's probably cleaner for the for loops to be written as `for (int i = channelNumbers.getStart(); i < channelNumbers.getEnd(); ++i)`
As that more clearly maps on to the inclusive start/exclusive end?

svenemtell:
That's fine with me!
Feel free to make the change and also change the comments to suit the change.
For the sake of conformity, the same change should be made in commit 6857dde.

drowaudio:
Should be fixed here now: https://github.com/Tracktion/tracktion_engine/commit/ac5fa7ce9cf19807ea12b3a89d36ce7706f90b64

PR #224, opened 2024-07-30
Feature/auv3

drowaudio:
- [x] Add function to ExternalPlugin to find out when a plugin is being created async
- [x] Add a way for the child process scanner to scan for AUv3s on a background thread to pick them up (can't be done on the message thread)
  - [x] Use `AudioFormat::findAllTypesForFile` then filter out the ones that `AudioFormat::requiresUnblockedMessageThreadDuringCreation` and run those on a background thread


julianstorer (commented):
This all looks pretty solid to me - can't see anything scary..

PR #245, opened 2025-01-18
Plugins: Add Automatable params to Chorus and Delay

botoxparty:
During the development of an application I discovered that some of these plugin's parameters do not have automatable parameters.

Followed the pattern from the reverb plugin.


drowaudio:
I think those parameters aren't automatable as they cause audible artefacts when adjusted?

botoxparty:
I haven't observed that.

I could imagine it for some cases, but for the mix?

PR #255, opened 2025-04-18
Fix playing notes notes retriggering when changing clip during first loop in Clip Launcher

dikadk:
Step to reproduce:

1. Add StepClip to ClipLauncher ClipSlot 0
2. Update notes using GUI

Issue only exists in the first loop if clip is played by ClipLauncher, so I had to modify StepSequencerDemo to include ClipLauncher to debug. 
 

https://github.com/user-attachments/assets/d8d77c67-a263-4452-9baf-5ecb70258e66

https://github.com/user-attachments/assets/5f64b2ce-d13e-4a5c-ac8d-68421edea190

This is updated StepSequencerDemo which uses ClipLuancher
[StepSequencerDemo.h.zip](https://github.com/user-attachments/files/19812721/StepSequencerDemo.h.zip)

### DIFF
--- a/modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
+++ b/modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
@@ -1028,7 +1028,7 @@ public:
             generator->setTime (0.0);
         }
 
-        return exhausted();
+        return !exhausted();
     }
 
     bool exhausted() override
@@ -1267,7 +1267,10 @@ public:
                                              midiSourceID,
                                              0.0,
                                              isPlaying);
-            shouldCreateMessagesForTime = true;
+            
+            // Only create messages at clip start, otherwise use the existing note state
+            // This prevents notes from retriggering when the step clip content is modified during playback
+            shouldCreateMessagesForTime = blockStartBeatRelativeToClip <= 0.00001_bd;
         }
 
         if (shouldCreateMessagesForTime)


drowaudio:
Thanks for working on this.
There's quite a few changes involved here, can you briefly go over the problem and rationale behind the fix please so I've got some background for the code review?

dikadk:
> Thanks for working on this. There's quite a few changes involved here, can you briefly go over the problem and rationale behind the fix please so I've got some background for the code review?

Yep, sure. 
Already playing notes were getting "stopped" and "started" when Clip's midi sequence was regenerated. It was happening because in LoopingMidiNode when sequencer was changing, it was clearing all playing notes and starting them in the same block without accounting if any of them where playing. Video of the issue is on the first video.
Then I came across this comment: https://github.com/Tracktion/tracktion_engine/blob/cd002bb9240d387c99168ca36229cc599a2838b0/modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp#L726

I tracked it down to only happening on the first loop of LoopingMidiNode and added logic on account of playing notes.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1170,10 +1194,13 @@ class GeneratorAndNoteList
             return;
 
         assert (sequences.size() > 0);
-        dynamicOffsetBeats = std::move (dynamicOffsetBeatsToUse);
-        shouldCreateMessagesForTime = clipPropertiesHaveChanged || noteListToUse == nullptr;
+        dynamicOffsetBeats = !dynamicOffsetBeatsToUse ? std::move (dynamicOffsetBeatsToUse)
```
drowaudio, line 1197: Why is this change needed?

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1170,10 +1194,13 @@ class GeneratorAndNoteList
...
+                                                          : std::make_shared<BeatDuration>();
         activeNoteList = noteListToUse ? std::move (noteListToUse)
                                        : std::make_shared<ActiveNoteList>();
+        
+        // If the clip properties have changed, we need to recreate messages
+        shouldCreateMessagesForTime = clipPropertiesHaveChanged || activeNoteList->areAnyNotesActive() == false;
```
drowaudio, line 1203: I'm not sure I understand this logic. Why would you want to create the messages for the current time if there are no active notes? Surely if there are active notes, you want to start them?

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1183,13 +1210,14 @@ class GeneratorAndNoteList
 
         sequencesHash = std::hash<std::vector<juce::MidiMessageSequence>>{} (sequences);
 
-        if (sequencesHash != lastSequencesHash || clipPropertiesHaveChanged)
-            shouldSendNoteOffsForNotesNoLongerPlaying = true;
+        // Force note-offs for all active notes if the sequence has changed
+        shouldSendNoteOffsForNotesNoLongerPlaying = (sequencesHash != lastSequencesHash) || clipPropertiesHaveChanged;
```
drowaudio, line 1214: Is this an intended change? This will always overwrite the `shouldSendNoteOffsForNotesNoLongerPlaying` rather than setting it to true only when the sequences/properties have changed.

I just want to double check this is required as part of the fix rather.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1262,12 +1290,25 @@ class GeneratorAndNoteList
         if (! isContiguousWithPreviousBlock
             || blockStartBeatRelativeToClip <= 0.00001_bd)
         {
-            MidiNodeHelpers::createNoteOffs (*activeNoteList,
-                                             destBuffer,
-                                             midiSourceID,
-                                             0.0,
-                                             isPlaying);
-            shouldCreateMessagesForTime = true;
+            // Only send note-offs when actually recreating the node - avoid turning off notes unnecessarily
```
drowaudio, line 1293: This change is a bit unclear as the `MidiNodeHelpers::createNoteOffs` call that it replaces also sends controller and all-notes-off messages. This new code doesn't seem to do any of that so I can imagine stuck notes when the sustain pedal is used.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1025,10 +1030,10 @@ class LoopedMidiEventGenerator : public MidiGenerator
         if (exhausted() && ! loopTimes.isEmpty())
         {
             setLoopIndex (loopIndex + 1);
-            generator->setTime (0.0);
+            generator->setTime (loopTimes.getStart());
         }
 
-        return exhausted();
+        return !exhausted();
```
drowaudio, line 1036: This looks like an important fix.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1025,10 +1030,10 @@ class LoopedMidiEventGenerator : public MidiGenerator
         if (exhausted() && ! loopTimes.isEmpty())
         {
             setLoopIndex (loopIndex + 1);
-            generator->setTime (0.0);
+            generator->setTime (loopTimes.getStart());
```
drowaudio, line 1033: I think this is correct but does it have a material effect on the bug?
I would have thought that the next call would either check the event time or set the time to the current position?

drowaudio (commented):
I'm trying to understand all the ins and outs here but there are lots of small changed that could have a big effect.

I guess what I'm struggling with is what bit is actually broken with the current implementation? The `activeNoteList` is supposed to track the active notes and then be checked when properties of the clip or sequence change.
IIRC the idea is that the current sequence is read and if a note-on is in the sequence but an "active note" is already found, it should be skipped.
If you're saying this isn't happening, then the fix should be much smaller than all these changes right?

Also, do you know why this only happens during the first loop?

I'm just very worried that in making this changes I break something else subtle.

Ideally, I'd need some unit tests but the MIDI is one of the areas I've yet to fully add tests for. Since I wrote this, I've written a testing audio device which can be used to run the audio engine in a test harness very precisely to check behaviour. I just need time to construct test cases.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
dikadk: Yeah, after retesting it, this change doesn't really impact fixing the bug. Reverted

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
dikadk: Oh, wasn't aware of possible issues with sustain pedal. I've revised code and only left single change that fixes bug of notes "re-triggering".

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
dikadk: Updated, not needed for a bug fix.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
dikadk: Updated, not needed for a bug fix

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
dikadk: Revised, not needed for a bug fix

dikadk:
@drowaudio Thanks for reviewing PR. 
> I'm trying to understand all the ins and outs here but there are lots of small changed that could have a big effect.
> 
> I guess what I'm struggling with is what bit is actually broken with the current implementation? The `activeNoteList` is supposed to track the active notes and then be checked when properties of the clip or sequence change. IIRC the idea is that the current sequence is read and if a note-on is in the sequence but an "active note" is already found, it should be skipped. If you're saying this isn't happening, then the fix should be much smaller than all these changes right?

 Yeah, you're right, fix is smaller. 

 I was able to scrape unnecessary changes I thought were needed to fix re-trigger bug. Especially bringing back handling of sustain pedal to prevent stuck notes. 
> I'm just very worried that in making this changes I break something else subtle.

> Ideally, I'd need some unit tests but the MIDI is one of the areas I've yet to fully add tests for. Since I wrote this, I've written a testing audio device which can be used to run the audio engine in a test harness very precisely to check behaviour. I just need time to construct test cases.

Nice, I would nice test this in the isolated environment.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -711,9 +711,13 @@ struct EventGenerator   : public MidiGenerator
                                 bool useMPEChannelMode, MPESourceID midiSourceID,
                                 juce::Array<juce::MidiMessage>& controllerMessagesScratchBuffer) override
     {
-        thread_local MidiMessageArray scratchBuffer, cleanedBufferToMerge;
+        ActiveNoteList originalState;
```
drowaudio, line 714: Ok, I think we're getting close now but I can't see why this `originalState` is required?
It looks like you're just making a copy of `activeNoteList`, using `originalState` but then updating `activeNoteList`?

Is it something to avoid modifying the `activeNoteList` state whilst iterating and checking it? (As the non-modified `originalState` is now being used?) If so, what was the problem this was solving?

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -723,32 +727,29 @@ struct EventGenerator   : public MidiGenerator
...
         {
             if (e.isNoteOn())
             {
-                if (! activeNoteList.isNoteActive (e.getChannel(), e.getNoteNumber()))
+                if (!originalState.isNoteActive(e.getChannel(), e.getNoteNumber()))
                 {
-                    cleanedBufferToMerge.add (e);
-                    activeNoteList.startNote (e.getChannel(), e.getNoteNumber());
+                    destBuffer.addMidiMessage(e, midiSourceID);
+                    activeNoteList.startNote(e.getChannel(), e.getNoteNumber());
```
drowaudio, line 737: I know it's a small thing but can you add the spaced beck after `! ` and before ` (` please?
I think that will just simplify the diff a bit as there's a few lines that are just missing this space.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1063,7 +1071,19 @@ class LoopedMidiEventGenerator : public MidiGenerator
 
         loopIndex = newLoopIndex;
         const auto sequenceOffset = clipRange.getStart() + (loopIndex * loopTimes.getLength());
-        generator->cacheSequence (sequenceOffset, loopTimes + sequenceOffset);
+        generator->cacheSequence(sequenceOffset, loopTimes + sequenceOffset);
```
drowaudio, line 1074: Again, I think this line is the same, just missing a space?

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1063,7 +1071,19 @@ class LoopedMidiEventGenerator : public MidiGenerator
...
+        ActiveNoteList notesAtLoopStart = generator->getNotesOnAtTime(loopTimes.getStart(),
+                                                  channelNumbers,
+                                                  clipLevel);
+
+        activeNoteList->reset();
+
+        // Track which notes we're preserving vs adding new
+        notesAtLoopStart.iterate([&](int channel, int note) {
+            activeNoteList->startNote(channel, note);
+        });
```
drowaudio, line 1086: So it looks like this is the most important part of the fix?
I'm not sure it's using the right variables though, `activeNoteList` is supposed to track the notes that have been started/stopped so resetting it without explicitly sending note-off events means it would get out of sync?

Are we sure that the only notes that are "active" are those `notesAtLoopStart`?
I think I'd prefer to make a change that turns off active notes then uses the `notesAtLoopStart` to turn on (and send) the correct note-ons?

The way you've done it just feels a bit brittle?

dikadk:
@drowaudio, I've revised PR and confirmed this bug is fixed even with these minimal changes. Sorry for making you go through some extra code. Your questions helped me narrow down the code changes needed.

modules/tracktion_engine/playback/graph/tracktion_LoopingMidiNode.cpp
```diff
@@ -1267,7 +1267,10 @@ class GeneratorAndNoteList
                                              midiSourceID,
                                              0.0,
                                              isPlaying);
-            shouldCreateMessagesForTime = true;
+            
+            // Only create messages at clip start, otherwise use the existing note state
+            // This prevents notes from retriggering when the step clip content is modified during playback
+            shouldCreateMessagesForTime = blockStartBeatRelativeToClip <= 0.00001_bd;
```
drowaudio, line 1273: Are you actually sure this is needed? The existing block around that line is this:
```
        if (! isContiguousWithPreviousBlock
            || blockStartBeatRelativeToClip <= 0.00001_bd)
        {
            MidiNodeHelpers::createNoteOffs (*activeNoteList,
                                             destBuffer,
                                             midiSourceID,
                                             0.0,
                                             isPlaying);
            shouldCreateMessagesForTime = true;
        }
```
So unless the playhead jumps, or loops round, it's the same as your modified line?

dikadk: I think you're completely right. I was able to reproduce the issue in Waveform 13 simply using Midi Clip with single note. 

https://github.com/user-attachments/assets/8882ff5d-c0b2-4cc6-beef-1eb8b9e75ebd

Well, talking about the fix. I don't think shouldCreateMessagesForTime = blockStartBeatRelativeToClip <= 0.00001_bd; is THE solution. I think it just masking a bug in active note state comparing code.

Let me know if you have any suggestion on fixing. I cannot seem to figure out where to make a change.

drowaudio:
Ok, sorry for the long delay on this, it just took me a while to get some head space and a release out so I could think about it.
I've basically implemented what you've suggested here now: https://github.com/Tracktion/tracktion_engine/commit/2daa8e847c4e71bfec01994e7c478ee4538e1121

Grab it and let me know if it does indeed fix your original issue.
Thanks for your patience 🙌

PR #382, opened 2026-08-17
Fix typo in nanorange operation

Nebula-Developer:
Fixed a typo in nanorange.hpp where `x.value` was accessed instead of `x.value_`


drowaudio:
I've removed the nanorange dependancy now as it doesn't appear to be actively maintained, a lot of the algorithms are in the standard now. (Also Tristan has a newer algorithms library, Flux that would be a better candidate).

PR #398, opened 2026-08-26
Audio: Added an inter-channel alignment estimate to the StereoFieldAnalyser - delay, polarity and an ambiguity guard

drowaudio:


drowaudio:
### Follow-up: phase alignment as a corrective processor

This PR only *measures* the inter-channel offset — `interChannelDelayMs`,
`interChannelDelaySamples` and `polarityInverted` are read-only, and the analyser
stays a pass-through observer.

Actually *fixing* an offset is deliberately out of scope here, because a corrector
has to delay or invert a channel. That breaks the pass-through contract of the
consumer this was built for (Waveform's Loudness Meter plugin) and its
bit-identical-audio test, so it belongs in its own plugin rather than in a meter:

- a Phase Alignment utility with its own UI and tests
- manual and automatic modes, the automatic one driven by the cross-correlation
  estimate added here
- worth building once the read-only estimate has proven itself in the meter

Noting it here so the intent behind the read-only API is on the record — the fields
are there to be reported to the user, not to drive correction inside the analyser.

One caveat for whoever picks it up: the estimate is deliberately refused when the
correlation peak is not prominent (a sustained tone correlates equally at every
multiple of its period), so `alignmentValid` has to be honoured before acting on a
delay — an auto-align mode must not "correct" a signal the analyser has said it
cannot resolve.

PR #412, opened 2026-09-11
Playback: Fixed proxy-file clips playing 2 samples late

drowaudio:
## Summary
Clips playing from a proxy file were 2 samples late against the timeline. Any clip with Clip FX is forced onto a proxy, so e.g. a copy of a clip with Clip FX > Invert phase didn't null against the original (Tracktion/waveform_beta#933). Flattening or rendering such a clip baked the 2 extra samples into the result.

## Root cause
Proxy playback (and frozen tracks) go through `WaveNode`, which runs the file through a `juce::LagrangeInterpolator` even at a 1:1 ratio. That interpolator has a base latency of 2 input samples, which `WaveNode` never compensated for. The non-proxy path (`WaveNodeRealTime` / `LagrangeResamplerReader`) already compensates, which is why only proxy clips were affected.

A second, smaller bug showed up in the test: `WaveNode::editTimeToFileSample` rounded with `(int64_t) (x + 0.5)`, which truncates towards zero for negative times. A block straddling the clip start then read from one sample too late and used a non-unity ratio for that block, smearing the first samples of the clip.

## Fix (`tracktion_WaveNode.cpp`)
- Read `getBaseLatency()` extra frames per block and feed the resampler from that far in, so its output lines up with `fileStart`.
- After a reset (construction, or leaving the clip), the skipped frames prime the resampler's history first, so the very first output frame is exact too.
- `editTimeToFileSample` uses `std::floor (x + 0.5)` so negative times round like positive ones.

## Regression test
`runSampleAlignmentTests` in `tracktion_WaveNode.test.cpp`, run for both `WaveNode` and `WaveNodeRealTime` across all test setups. It plays a file with impulses at frame 0 and 1000, starting on a block boundary, and checks each lands on exactly the right timeline sample with silence either side.
- Before: 108 failures, all `WaveNode` (impulses at +2, and smeared in the random-block-size setups).
- After the latency fix only: 42 failures, all random-block-size setups (the rounding bug).
- After both: 4080/4080 pass. `WaveNodeRealTime` passed throughout.

Full TestRunner (Release, macOS): 345/345 test cases, 20835 assertions pass.

## Not changed (follow-ups)
- `SpeedRampWaveNode` (proxy clips with speed-ramp fades) is a copy of the same code and has both bugs. It's left alone here to keep this change small.
- Volume, Fade in/out, Step volume, Pitch shift and Plugin Clip FX render through `WaveNode` too, so their cached renders carry the same 2-sample shift. New renders are correct. Renders already cached keep the shift until something invalidates their hash. Invert, Normalise, Reverse and Make mono don't render through `WaveNode`, so the case reported in the issue is fixed without a re-render.

### DIFF
--- a/modules/tracktion_engine/playback/graph/tracktion_SpeedRampWaveNode.cpp
+++ b/modules/tracktion_engine/playback/graph/tracktion_SpeedRampWaveNode.cpp
@@ -19,6 +19,7 @@ struct SpeedRampWaveNode::PerChannelState
 
     juce::LagrangeInterpolator resampler;
     float lastSample = 0;
+    bool resamplerNeedsPriming = true;
 };
 
 
@@ -147,8 +148,10 @@ int64_t SpeedRampWaveNode::editTimeToFileSample (TimePosition editTime) const no
         jassert (speedFadeDescription.outTimeRange.containsInclusive (editTime));
     }
 
-    return (int64_t) ((editTime - (editPosition.getStart() - offset)).inSeconds()
-                       * originalSpeedRatio * audioFileSampleRate + 0.5);
+    // N.B. floor rather than truncate so times before the clip start (which happen when a
+    // block straddles it) round to the nearest sample like all the others
+    return (int64_t) std::floor ((editTime - (editPosition.getStart() - offset)).inSeconds()
+                                  * originalSpeedRatio * audioFileSampleRate + 0.5);
 }
 
 bool SpeedRampWaveNode::updateFileSampleRate()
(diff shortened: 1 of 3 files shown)


drowaudio:
We need to apply the same fix to speedrampwavenode

drowaudio:
🤖 Done in b9454cac1d0 - SpeedRampWaveNode had its own copy of the same code, so it got the same three changes: read the resampler's base latency worth of extra frames and feed it from that far in, prime its history after a reset, and floor rather than truncate in editTimeToFileSample.

The alignment test now covers that node too. It skips the randomised block sizes because editTimeToFileSample clamps a block straddling the clip start into the clip, so that block resamples at a non-unity ratio and can't line up to the sample, and it places its impulses past the largest block size because the node ramps the gain up across the first block it plays. With the fix reverted the new checks fail on all 24 fixed-block-size setups (impulses 2 samples late); with it they pass, and the full TestRunner is green (345/345 test cases, 20955 assertions).

PR #427, opened 2026-09-15
Control Surfaces: Pads can show the host's clip colours and refresh on track colour changes

drowaudio:
## Summary
Control surfaces can now show clip launcher pads in the colours the host UI draws, and pads refresh when a track's colour changes.

## Root cause
- `ExternalController::updatePadColours()` only had `Clip::getColour()`, quantised into a hard-coded 18-slot index. Hosts that draw clips differently had no way to tell surfaces the real colour. That includes a user palette, and Waveform's link-to-track-colour setting.
- `ExternalControllerManager`'s edit watcher never refreshed pads on track colour changes.

## Changes
- **`UIBehaviour` hooks:**
  - `getClipColourForControlSurface (const Clip&)` returns the drawn colour. The default is `Clip::getColour()`.
  - `getClipColourIndexForControlSurface (const Clip&)` returns 0 or 1 to N. The default quantises the hue of the colour hook.
  - `getNumClipColourIndexes()` defaults to 18.
- **Pad callback:** `ControlSurface::padColourStateChanged (channel, scene, colourIdx, juce::Colour, state)` is now what `updatePadColours()`/`clearPadColours()` call. It defaults to `padStateChanged()`, so existing surfaces are unchanged. The colour is transparent for empty, playing, queued and recording pads.
- **Track colour refresh:** a `TRACK` `IDs::colour` change triggers a pad update.
- **Changelist:** `CHANGELIST.md` entry under v3.5 > Misc.

## Tests
`control_surfaces/tracktion_ExternalController.test.cpp`:
- The default index quantisation and the count come from `UIBehaviour`.
- The default colour hook returns the clip's own colour.
- An overridden colour hook drives the default index.
- `padColourStateChanged` forwards to `padStateChanged` by default.

The full TestRunner passes (355/355).

App PR: Tracktion/waveform#121

Fixes Tracktion/waveform_beta#1260


drowaudio:
On the engine side I think the number of indices will need to be returned via UIBehaviour as I think the 18 colour palette is specific to Waveform, not the engine.

Also, the index doesn't relate to hue anymore as these can be set individually. Is that supported?

drowaudio:
🤖 Good points, thanks.

**1. The colour count belongs in UIBehaviour.** Agreed, and I think the whole index calculation should move there, not just the count. Proposed revision:
- Replace `getClipColourForControlSurface()` with two hooks:
  - `virtual int getNumClipColourIndexes()`, defaulting to 18.
  - `virtual int getClipColourIndexForControlSurface (const Clip&)`. It returns 0 for no colour, otherwise 1 to N. The default is the current hue-based calculation using the count above, so other hosts see no change.
- `updatePadColours()` calls the index hook. `limitedPadColours` and the playing/queued overrides stay in the engine.
- Waveform overrides the index hook with its own slot logic: the link-to-track rule, plus the same rounding as `getSkewedColour()`. Today the engine rounds slightly differently at the top of the hue range (a hue just under 1.0 is slot 18 in the engine but red in Waveform), and this removes that mismatch.
- `ExternalController::getPadColourIndex()` and its test go away, replaced by a test of the default hook.

**2. Colours set individually.** Two parts to this:
- **Slot index: supported.** Clips still store their palette slot encoded as a hue (`i / 18`), and Waveform maps that hue back to a slot in `getSkewedColour()`. The index the scripts receive is still the right slot. With the change above, Waveform computes it directly, so the engine no longer assumes anything about hue.
- **Custom colour schemes: not supported.** `ColourManager::Props` allows each of the 18 colours to be overridden per scheme. The LED tables in the scripts are fixed, so they only match the default scheme. If a scheme recolours a slot, its pads keep the default colour. Following custom schemes would need the actual colour sent to the scripts, i.e. the RGB approach we decided against.

Shall I go ahead with the revision in 1, and accept the limitation in 2 (pads match the default scheme only)?

drowaudio:
🤖 Implemented the revision:
- **New hooks:** `UIBehaviour::getNumClipColourIndexes()` (default 18) and `getClipColourIndexForControlSurface (const Clip&)` (the default quantises hue with that count, same result as before). They replace `getClipColourForControlSurface()`.
- **Pad code:** `updatePadColours()` uses the index hook, and `ExternalController::getPadColourIndex()` is removed from the public API.
- **Test:** covers the default mapping and a UIBehaviour with a different colour count. I dropped the transparent-colour assertion, because `Clip::setColour()` resets a transparent colour to the clip's default, so a clip never reports transparent.
- **Results:** the full TestRunner passes (354/354). The app side (Tracktion/waveform#121) is updated to the new hooks and builds.

drowaudio:
Update the change list for this

drowaudio:
🤖 Added a line to `CHANGELIST.md` under v3.5 > Misc (commit aa6da9f79f0):

> New `UIBehaviour::getNumClipColourIndexes()` and `UIBehaviour::getClipColourIndexForControlSurface()` so hosts can match control surface clip pad colours to their own palette, and pads now refresh when a track's colour changes

I didn't add a `BREAKING-CHANGES.md` entry. The change only adds virtuals with defaults, and nothing on develop was removed or changed. The app submodule on Tracktion/waveform#121 now points at this commit.

PR #430, opened 2026-09-16
Clips: Split clip creation defaults from clip moves and copies

drowaudio:
Separates creating a clip from moving or copying one, so a clip that already exists keeps the settings the user gave it.

`insertClipWithState` applied the defaults for a new clip on every insert, so pasting, splitting or dragging an existing clip re-applied them. For a clip in a `ClipSlot` that meant its loop and tempo-remap settings were reset every time it was copied: Tracktion/waveform_beta#1032. The app side is in Tracktion/waveform#123.

## The split

- **`insertClipWithState`** is the creation route: state defaults → insert → clip defaults → `newClipCreated` → launcher preparation when the owner is a `ClipSlot`. Signature unchanged, and it's what `insertNewClip`, `insertWaveClip`, `insertMIDIClip`, `insertEditClip` and `createClipFromFile` funnel through.
- **`insertClipCopy (ClipOwner&, const ClipCopy&)`** is the transfer route. `ClipCopy` can only be built from a live clip (`fromClip`) or a clipboard state (`fromClipboardState`), and carries whether the source was in the launcher. `withNewItemID (Edit&)` covers copying a clip that stays in the Edit. It applies only the fill-in-if-missing state defaults and the track-hue colour, so nothing the user set is overwritten.
- **`prepareClipForLauncher (Clip&)`** is now one function instead of being copied into the clipboard, the recording paths and the app. It leaves a one-shot unlooped.
- **`Clip::moveTo`** prepares a clip that moves in from the arrangement.

Now routed through the transfer path: clipboard clip paste (launcher or arrangement source), marker/chord/arranger paste, scene paste, split, `duplicateClip` and the preview Edit.

## EngineBehaviour

`newClipCreated` is called once per created clip. `newClipAdded` is deprecated and now only called alongside it, so it no longer fires for moves, paste or split. Both BREAKING-CHANGES.md entries cover this and `duplicateClip` now respecting `maxClipsInTrack`.

## Also removed

The clipboard's flush of `proxyAllowed`/`resamplingQuality` onto copied states, the duplicated launcher blocks in `Clipboard.cpp` and the two input devices, and a dead `IDs::stretchMode` write (only read alongside the legacy `timeStretch` property).

## Tests

New `model/clips/tracktion_ClipDefaults.test.cpp` behind `ENGINE_UNIT_TESTS_CLIP_DEFAULTS`, covering: creation defaults on a track and in a slot (audio and MIDI), a move between tracks keeping its settings, a launcher clip copied or pasted into another slot keeping loop/auto-tempo/sync, colour handling on copies, one-shots not being looped, and the hook firing once per creation and not on a move.

Full suite: 368 test cases, 21240 assertions, all passing.

## Notes

- `Clip::moveTo (ClipSlot&)` used to return `false` even when it moved the clip. That's fixed, but it's unreachable today: `canContainAudio`/`canContainMIDI` only accept a `Track`, so `canBeAddedTo` rejects a `ClipSlot` for both audio and MIDI clips. Worth deciding separately whether slots should be accepted there.


modules/tracktion_engine/model/clips/tracktion_ClipOwner.cpp
```diff
@@ -262,179 +263,279 @@ namespace clip_owner
...
+{
+    return { clip.state.createCopy(), clip.getClipSlot() != nullptr };
+}
+
+ClipCopy ClipCopy::fromClipboardState (juce::ValueTree stateToUse, bool sourceWasInLauncher)
+{
+    return { std::move (stateToUse), sourceWasInLauncher };
+}
+
+ClipCopy ClipCopy::withNewItemID (Edit& edit) const
```
drowaudio, line 492: **`withNewItemID` only rewrites the root ID and automation curve modifier IDs.** It doesn't go through `EditItemID::remapIDs`, so nested ID-bearing children keep the source clip's IDs: plugin states inside `ClipEffects` (`tracktion_ClipEffects.cpp` `PluginEffect`), and child clips of a `ContainerClip`. A same-edit copy made with `ClipCopy::fromClip (c).withNewItemID (edit)` then has two plugins sharing one `EditItemID`, and `PluginCache::getPluginFor (EditItemID)` returns whichever it finds first. `split` has had the same latent issue for a while, but this turns it into a public API whose doc ("for when the clip it was taken from is staying in the Edit") reads as if it were safe. Its only caller today is the test helper `insertCopyOfClip`. Suggest calling `EditItemID::remapIDs (newState, nullptr, edit)` here, the same as `duplicateClip` and the clipboard paste.

drowaudio (commented):
Review of db027cf1593 (all three commits). Not a blocking review; nothing was pushed.

I built the TestRunner on this branch and ran the full doctest suite: 366/366 test cases pass, 11724 assertions, no expected-fail assertions. The six new `ClipDefaults` cases all pass unconditionally.

Ten inline comments, roughly in severity order:

1. `ClipCopy::withNewItemID` doesn't remap nested IDs (clip-effect plugins, container clip children) the way `EditItemID::remapIDs` does.
2. Marker/chord/arranger paste still goes through `insertClipWithState`, so paste fires `newClipCreated` for those clip types.
3. `duplicateClip`'s launcher flag is dead because the target is always the host `ClipTrack`; launcher duplicates land on the arrangement (pre-existing), so waveform_beta#1032 may still reproduce if the app duplicates via `duplicateClip`.
4. Copies skip the creation fill-ins (track-hue colour, `sync`, `autoCrossfade`, `LOOPINFO`), which is observable for clips from older edits.
5. `newClipAdded` semantics narrowed and reordered with no BREAKING-CHANGES.md entry.
6. `duplicateClip` can now return null and show a modal warning at `maxClipsInTrack`.
7. Stale `should_fail` comment in the test file, and the PR description still says "tests only".
8. The named `Engine` in the hook test nulls `Engine::instance` and writes a Settings.xml under ~/Library (mirrors the Archive test).
9. Design: `prepareClipForLauncher` overwrites explicit settings, forcing the `sourceWasInLauncher` plumbing through five call sites.
10. Minor: `insertClipWithState` duplicates `insertClipCopy`'s tail.

Two things I checked and consider fine: `Clip::moveTo` now returning `true` for a slot move fixes a pre-existing fall-through to `false`, and dropping `IDs::stretchMode` from the creation defaults is a no-op since `AudioClipBase` only reads it when the legacy `timeStretch` property is set.

modules/tracktion_engine/selection/tracktion_Clipboard.cpp
```diff
@@ -1135,13 +1109,15 @@ bool Clipboard::Clips::pasteIntoEdit (const EditPastingOptions& options) const
...
-                    if (auto newClip = insertClipWithState (*clipSlot, newClipState))
+                    if (auto newClip = insertClipCopy (*clipSlot, ClipCopy::fromClipboardState (newClipState,
+                                                                                                clip.slotOffset.has_value())))
                         itemsAdded.add (newClip);
                 }
             }
             else if (auto clipTrack = dynamic_cast<ClipTrack*> (targetTrack->getSiblingTrack (clip.trackOffset, false)))
             {
-                if (auto newClip = clipTrack->insertClipWithState (newClipState))
+                if (auto newClip = insertClipCopy (*clipTrack, ClipCopy::fromClipboardState (newClipState,
```
drowaudio, line 1119: **The marker, chord and arranger branches above (lines 1044, 1061, 1069) still go through `insertClipWithState`.** Those are pasted copies of existing clips, but `insertClipWithState` now fires `EngineBehaviour::newClipCreated` (and the deprecated `newClipAdded`) unconditionally, so pasting a marker/chord/arranger clip reports a created clip. That contradicts the new doc on `newClipCreated` ("It isn't called when an existing clip is moved, copied, pasted or split"). They should use `insertClipCopy` like this branch; the creation defaults are no-ops for those clip types anyway.

modules/tracktion_engine/model/edit/tracktion_EditUtilities.cpp
```diff
@@ -326,16 +326,16 @@ Clip::Ptr duplicateClip (const Clip& c)
...
     jassert (c.edit.clipCache.findItem (newClipID) == nullptr);
 
     if (auto t = c.getClipTrack())
     {
         jassert (! t->state.getChildWithProperty (IDs::id, newClipID).isValid());
-        t->state.appendChild (n, c.getUndoManager());
 
-        if (auto newClip = t->findClipForID (newClipID))
+        // The copy is of an existing clip, so it keeps that clip's settings
+        if (auto newClip = insertClipCopy (*t, ClipCopy::fromClipboardState (n, c.getClipSlot() != nullptr)))
```
drowaudio: **The `c.getClipSlot() != nullptr` flag is dead here, and the launcher duplicate case may still be unfixed.** `t` comes from `c.getClipTrack()`, and `Clip::getTrack()` falls back to the slot's host track for a launcher clip, so the target is always the arrangement `ClipTrack`. `insertClipCopy` only consults `wasInLauncher()` when the owner is a `ClipSlot`, so the flag never has an effect, and a duplicated launcher clip lands on the arrangement track at start 0 (pre-existing, the old `appendChild` did the same). If Waveform's launcher duplicate action calls `duplicateClip`, this commit doesn't fix waveform_beta#1032 for that route; only the clipboard and `insertClipCopy` routes are covered by the new tests. Worth either routing slot clips into a free slot here or dropping the flag and documenting that `duplicateClip` targets the track.

modules/tracktion_engine/model/clips/tracktion_ClipOwner.cpp
```diff
@@ -262,179 +263,279 @@ namespace clip_owner
...
+{
+    auto newState = state.createCopy();
+    edit.createNewItemID().writeID (newState, nullptr);
+    assignNewIDsToAutomationCurveModifiers (edit, newState);
+
+    return { std::move (newState), sourceWasInLauncher };
+}
+
+Clip* insertClipCopy (ClipOwner& clipOwner, const ClipCopy& clipCopy)
```
drowaudio, line 501: **Copies no longer get any of the creation-time fill-ins.** Paste, split, `Scenes::pasteIntoEdit` and `createEditForPreviewingClip` used to go through `insertClipWithState`, which recoloured a clip with the default colour to the destination track's hue, and filled in `sync`, `autoCrossfade` and `LOOPINFO` when missing. `Clip::colour` has no explicit `CachedValue` default, so a clip from an older edit without a `colour` property, pasted onto another track, now keeps the default colour instead of picking up the track hue. Missing `sync` falls back to `syncBarsBeats`, which differs from the `EngineBehaviour` remap settings. This may be an acceptable consequence of "copies keep their settings", but nothing in the new tests pins colour or sync on a copy, so it's worth deciding deliberately. One option: apply `applyCreationStateDefaults` (the missing-property fill-ins only) in `insertClipCopy` too.

modules/tracktion_engine/utilities/tracktion_EngineBehaviour.h
```diff
@@ -265,7 +265,15 @@ class EngineBehaviour
...
 
+    /// Allows a newly created clip to be customised.
+    /// This is called once, when a clip is created, after the engine's own
+    /// defaults have been applied and before a clip in a ClipSlot is prepared for
+    /// the launcher. It isn't called when an existing clip is moved, copied,
+    /// pasted or split.
+    virtual void newClipCreated (Clip&, [[ maybe_unused ]] bool fromRecording)      {}
+
     /// Allows a new clip to be customised.
+    [[deprecated ("Use newClipCreated instead")]]
```
drowaudio: **`newClipAdded` silently changes meaning, and there's no BREAKING-CHANGES.md entry.** Before, it fired from `ClipList::newObjectAdded` for every clip object added outside undo/redo: moves via `ClipSlot::setClip`, `duplicateClip`, clipboard paste and split. Now it fires only from `insertClipWithState`, and after `applyCreationClipDefaults` rather than during `addChild` (so an override that checks `getColour() == getDefaultColour()` now sees the engine's colour). An app overriding it compiles clean (overriding a deprecated virtual doesn't warn) and just stops getting callbacks. The deprecation message only says "Use newClipCreated instead". Suggest a BREAKING-CHANGES.md entry describing both the narrowing and the ordering change, and mentioning it in the attribute text. (Note: `rebuildObjects` never called `newObjectAdded`, so Edit load was never affected.)

modules/tracktion_engine/model/clips/tracktion_ClipOwner.cpp
```diff
@@ -262,179 +263,279 @@ namespace clip_owner
...
     {
+        CRASH_TRACER
+        jassert (clipState.isValid());
+        jassert (! clipState.getParent().isValid());
+
+        auto& edit = clipOwner.getClipOwnerEdit();
+
+        if (clipOwner.getClips().size() >= edit.engine.getEngineBehaviour().getEditLimits().maxClipsInTrack)
+        {
+            edit.engine.getUIBehaviour().showWarningMessage (TRANS("Can't add any more clips to this track!"));
```
drowaudio, line 348: **`duplicateClip` can now fail and pop a modal warning.** It used to `appendChild` unconditionally; routing it through `addClipState` adds the `maxClipsInTrack` check, which returns `nullptr` and calls `UIBehaviour::showWarningMessage` from inside a model-layer utility. `duplicateClip` is public API with no in-tree callers, so any host code that dereferences its result or loops "duplicate N times" now hits a null plus an unsolicited dialog at the limit. Probably fine, but worth calling out in the PR description / BREAKING-CHANGES.md.

modules/tracktion_engine/model/clips/tracktion_ClipDefaults.test.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+namespace tracktion::inline engine
+{
+
+//==============================================================================
+/** Characterisation tests for the defaults applied to clips.
+
+    These pin down which defaults are applied when a clip is created, and which
+    of a clip's settings survive being moved or copied. Cases marked
+    should_fail() describe the behaviour we want but don't have yet - they flip
```
drowaudio: **Stale comment: nothing in this file uses `should_fail()`.** The `should_fail` decorators were removed in the second commit when the behaviour change landed, and the full suite runs with 366/366 passing and 0 expected-fail assertions. The PR description also still says "tests only, no behaviour change" and describes the three cases as expected failures. Both should be updated to describe the PR as it is now, since the description currently undersells the scope (ClipOwner, Clip::moveTo, Clipboard, EditUtilities, the two input devices and the EngineBehaviour deprecation).

modules/tracktion_engine/model/clips/tracktion_ClipDefaults.test.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        // newClipCreated is called from the creation route only, so moves and
+        // copies aren't reported to the app as new clips.
+        // NB: this temporarily becomes Engine::instance; safe here because the
+        // engine TestRunner never calls Engine::getInstance() and other tests use
+        // getEngines().
+        auto behaviour = std::make_unique<CountingEngineBehaviour>();
+        auto behaviourPtr = behaviour.get();
+
+        Engine engine { juce::String ("tracktion_clip_defaults_test"), nullptr, std::move (behaviour) };
```
drowaudio, line 355: **Constructing a named `Engine` here has two side effects.** `Engine::Engine` sets the static `instance = this` and `~Engine` sets it to `nullptr` without restoring the previous value (`tracktion_Engine.cpp:22` and `:137`), so `Engine::getInstance()` is null for the rest of the run (harmless today because `TRACKTION_ENABLE_SINGLETONS` is off, as the NB says). It also writes `~/Library/tracktion_clip_defaults_test/Settings.xml` on the developer's machine via `PropertyStorage`. This mirrors the existing pattern in `tracktion_Archive.test.cpp:3543`, so not new, but a shared test helper that installs a custom `EngineBehaviour` (and ideally uses a temporary settings location) would stop this being copied a third time.

modules/tracktion_engine/model/clips/tracktion_ClipOwner.cpp
```diff
@@ -262,179 +263,279 @@ namespace clip_owner
...
+        return dynamic_cast<ClipSlot*> (clipOwner.getClipOwnerSelectable()) != nullptr;
+    }
+}
 
-                    if (! sc->isLooping())
-                        sc->setLoopRangeBeats ({ 0_bp, sc->getLengthInBeats() });
-                }
-            }
+//==============================================================================
+void prepareClipForLauncher (Clip& clip)
```
drowaudio, line 443: **Design note: `prepareClipForLauncher` overwrites explicit settings, so every caller has to remember a "was already in the launcher" flag.** It unconditionally does `setAutoTempo (true)`, `setUsesProxy (false)` and `setStart (0)`, which is why `ClipCopy` needs `sourceWasInLauncher`, `Clipboard::ClipInfo::slotOffset` doubles as that flag, `Scenes::pasteIntoEdit` hardcodes `true`, and `duplicateClip` passes a value that has no effect. Five call sites now have to get the flag right (moveTo, insertClipWithState, insertClipCopy, MidiInputDevice, WaveInputDevice), and `addClipState` reimplements `ClipSlot::setClip`'s "evict the existing clip and add the child" instead of calling it, so there's no single funnel into a slot. Two more general alternatives: make this function only fill in settings that aren't already explicit on the state (the same `setPropertyIfMissing` style used by `applyCreationStateDefaults`), or route every slot insertion through `ClipSlot::setClip` and prepare there once. Either would remove the bool plumbing.

modules/tracktion_engine/model/clips/tracktion_ClipOwner.cpp
```diff
@@ -262,179 +263,279 @@ namespace clip_owner
...
+            prepareClipForLauncher (*newClip);
+
+        return newClip;
+    }
+
+    return {};
+}
+
+//==============================================================================
+Clip* insertClipWithState (ClipOwner& clipOwner, juce::ValueTree clipState)
```
drowaudio, line 527: **Minor: this duplicates the tail of `insertClipCopy`.** Both end with `addClipState` then `if (isClipSlot (clipOwner)) prepareClipForLauncher (*newClip)`, and they already differ slightly (`insertClipCopy` also checks `! wasInLauncher()`). After `applyCreationStateDefaults`, this could call `insertClipCopy (clipOwner, ClipCopy::fromClipboardState (clipState, false))` and then apply `applyCreationClipDefaults` plus the callbacks, so the launcher preparation lives in one place.

drowaudio:
🤖 Thanks - I've checked all ten against the code. Eight I agree with, one I'd like to push back on, one I'd defer. Nothing pushed yet; say the word and I'll do the agreed set in one commit.

**Agreed, will fix**

1. **`withNewItemID` nested IDs** - correct, it only writes the root ID and the automation curve modifiers. Clip-effect plugin states and `ContainerClip` children keep the source IDs. It'll call `EditItemID::remapIDs (newState, nullptr, edit)` like `duplicateClip` and the clipboard do. Its only caller today is the test helper, so nothing in-tree is affected, but the doc makes it look safe and it isn't.

2. **Marker/chord/arranger paste** - confirmed at `Clipboard.cpp:1044/1061/1069`. They're pasted copies going through the creation route, so they fire `newClipCreated`, which contradicts the doc I wrote. They'll use `insertClipCopy`.

4. **Creation fill-ins on copies** - agreed this needs deciding rather than falling out. A clip from an older edit with no `colour` property keeps the default instead of the track hue, and missing `sync` now falls back to `syncBarsBeats` rather than the `EngineBehaviour` setting. Plan: `insertClipCopy` applies `applyCreationStateDefaults` (the "if missing" fill-ins only, which never override explicit values) plus the colour-if-still-default step, and the tests pin colour and sync on a copy. That keeps "a copy keeps what the user set" while still filling genuine gaps.

5 & 6. **`newClipAdded` and `duplicateClip`** - agreed, both go in BREAKING-CHANGES.md: the hook's narrowing *and* its new ordering (after `applyCreationClipDefaults`, so an override checking `getColour() == getDefaultColour()` sees the engine's colour), and `duplicateClip` now being subject to `maxClipsInTrack`, so it can return null and show a warning. I'll put the narrowing in the deprecation text too.

7. **Stale comment and PR description** - yes. The description is from the first commit; I'll rewrite it to cover the whole branch.

10. **Duplicated tail** - agreed, `insertClipWithState` will go through `insertClipCopy` for the insert plus launcher step.

**Agreed, with a caveat**

3. **`duplicateClip`'s flag is dead** - confirmed: `getClipTrack()` resolves through `getTrack()`, which falls back to the slot's host track (`tracktion_Clip.cpp:331-338`), so the owner is never a `ClipSlot` and the flag can't fire. I'll drop it and document that `duplicateClip` targets the track. On the issue itself: Waveform's duplicate is Cmd+D, which is copy + paste through `Clipboard::Clips` (`AcceleratorFunctions.cpp:307`), not `duplicateClip`, so waveform_beta#1032 is covered by the clipboard route and by the app changes in Tracktion/waveform#123. Routing slot clips into a free slot inside `duplicateClip` would be a behaviour change for existing API users, so I'd rather not do it here.

**Push back**

9. **"Only fill in settings that aren't already explicit"** - this is the option I'd avoid. An arrangement clip that the user explicitly un-looped, or explicitly set to no auto-tempo, would then arrive in a slot still un-looped and not beat-based, which is exactly the case the launcher defaults exist for and which we confirmed should keep working. The flag isn't incidental complexity: whether to convert depends on where the clip *came from*, which isn't in the state. Routing everything through `ClipSlot::setClip` doesn't remove it either, since a slot-to-slot move also goes through `setClip` and must not convert. What I can do is narrow the surface: have `addClipState` use `ClipSlot::setClip` for the evict-and-add step so there's one funnel into a slot, and keep the source flag as the only input. If you'd still prefer the fill-in-only semantics, I'll do it, but it changes the arrangement-to-launcher behaviour you confirmed earlier.

**Defer**

8. **Named `Engine` in the hook test** - agreed on both counts (static `instance` left null, `~/Library/tracktion_clip_defaults_test/Settings.xml`). It mirrors `tracktion_Archive.test.cpp:3543`, and a shared helper that installs a custom `EngineBehaviour` with a temporary settings path would stop it being copied a third time. That's a test-infrastructure change touching an existing test, so I'd rather do it as its own PR than fold it into this one. Happy to do it now if you'd prefer.

Shall I go ahead with 1, 2, 3 (drop the flag), 4, 5, 6, 7 and 10?

drowaudio:
>Shall I go ahead with 1, 2, 3 (drop the flag), 4, 5, 6, 7 and 10?

Yes

For 9, the edge case is clips that genuinely shouldn't be looped like single-hit drums or non-loopable recordings. If there isn't a robust way of telling if a clip has loop-disabled, maybe we could look at the source file and see if it's a one-shot (LoopInfo::isLoopable()/isOneShot()). If it is, then don't loop it.

drowaudio:
🤖 All eight are in `bdc9d67d01c`, plus the one-shot rule for 9.

1. `withNewItemID` uses `EditItemID::remapIDs`, so clip-effect plugins and container children get new IDs too.
2. Marker, chord and arranger paste use `insertClipCopy`.
3. `duplicateClip` passes `false` with a comment that it always targets the clip's track, since `getClipTrack` resolves through the slot's host track.
4. `insertClipCopy` now applies `applyCreationStateDefaults` (fill-in-if-missing only) and the track-hue colour when the clip still has the default. Two new cases pin it: an explicit colour survives a copy, and a state with no `colour` property gets the track hue.
5 & 6. Both BREAKING-CHANGES.md entries added: the `newClipAdded` narrowing *and* its new ordering, and `duplicateClip` now respecting `maxClipsInTrack` (null plus a warning at the limit). The deprecation message spells out both too.
7. Stale comment gone. PR title and description rewritten to describe the whole branch.
10. `insertClipWithState` and `insertClipCopy` share the colour step; the launcher preparation stays explicit in each because the creation route applies it unconditionally and the copy route only for a non-launcher source. Folding one into the other would mean passing a "don't prepare" flag, which seemed worse than the two-line tail.

**9 - one-shots.** Good idea, and simpler than the alternatives: `prepareClipForLauncher` now skips the loop range when `AudioClipBase::getLoopInfo().isOneShot()`, so a single hit stays unlooped while everything else still converts. New test covers it. Auto-tempo and start-at-0 still apply, since a one-shot still needs to be beat-aligned when launched. The `sourceWasInLauncher` flag stays for the "clip is already a launcher clip" case, which one-shot detection doesn't cover.

**Something that came out of the one-shot test:** `Clip::moveTo (ClipSlot&)` can't currently succeed at all. `canContainAudio` and `canContainMIDI` only accept a `Track`, so `canBeAddedTo` rejects a `ClipSlot` for both audio and MIDI clips (`tracktion_ClipOwner.cpp:941` and `:932`). The return-value fix from the earlier commit is therefore unreachable, and the app moves clips into slots by removing them and inserting instead. I left it alone since it's outside the agreed list and would widen `moveTo`'s behaviour. Want a follow-up that accepts a `ClipSlot` there?

**Verification:** engine 368 test cases, 21240 assertions, all green. Waveform14 rebuilt against this commit, and its unit tests pass (5 doctest cases, 143 assertions, no JUCE-category failures). Tracktion/waveform#123's submodule points here.

Left alone as agreed: the shared test-Engine helper (8) - happy to do that as its own PR.

drowaudio:
Yes, follow up with the Clip::moveTo.
I think canContainMIDI and canContainAudio need to get the track from the ClipSlot and call themselves?
Check this won't affect current behaviour that relies on that currently failing.

drowaudio:
🤖 Done in `a4690db0f0c`, exactly as you suggested: both predicates resolve a `ClipSlot` to the track it belongs to and ask themselves the same question.

```cpp
if (auto cs = dynamic_cast<const ClipSlot*> (&co))
    if (auto trackOwner = dynamic_cast<const ClipOwner*> (&cs->track))
        return canContainAudio (*trackOwner);
```

The extra cast is because `Track` isn't a `ClipOwner` - `ClipTrack` is - so the recursion goes through the track as an owner rather than taking `Track&` directly.

**What relies on the current failure: nothing that I can find.** The two `ClipOwner`-taking predicates are only reached from `AudioClipBase::canBeAddedTo`, `MidiClip::canBeAddedTo` and `StepClip::canBeAddedTo`, and the only caller of those is `Clip::moveTo` plus `DragClipsOperation::getDestinationTrack` in the app. That one does `dynamic_cast<ClipOwner*> (destTrack)` over `getAllTracks`, so it only ever asks about a `Track` and is unaffected. Marker, chord and arranger clips have their own `canBeAddedTo` and don't use these.

`Clip::moveTo (ClipSlot&)` therefore works now, and returns `true` when it does. A clip moved in from the arrangement is prepared for the launcher; one moved between slots keeps its own settings.

New test case covers all three: audio from the arrangement, MIDI from the arrangement, and a slot-to-slot move keeping loop and auto-tempo off.

**Verification:** engine 369 test cases, 21256 assertions, all green. Waveform14 rebuilt against this commit and its unit tests pass (5 doctest cases, 143 assertions, no JUCE-category failures). Tracktion/waveform#123's submodule points here.

Nothing in the app switches to `moveTo` for launcher drops in this PR - it still removes and inserts, which keeps the multi-clip offset handling - but the API is now usable for it if we want that later.

PR #437, opened 2026-09-22
Project: Freed each Edit's plugin instances during convertToFolderBasedProject

drowaudio:
## Summary
`convertToFolderBasedProject()` now calls `cleanUpDanglingPlugins()` after each Edit it converts, so that Edit's plugin instances are freed straight away. It asserts that it is running on the message thread, and only flushes when it is.

## Root cause
`ExternalPlugin` hands its instance to `AsyncPluginDeleter`, which frees one instance per 100 ms timer tick on the message thread. The conversion loop blocks the message thread, so nothing is freed until it returns. Every Edit's plugins, and all the memory they hold, pile up for the whole conversion. In Tracktion/waveform_beta#1274, memory grew by about 700 MB per sampler Edit until the machine locked up.

## Tests
The regression test lives in Waveform (`plugins/internal/MultiSamplerExamining.test.cpp`, "Converting a project frees each Edit's plugin instances as it goes"). An engine-only TestRunner never instantiates external plugins in examined Edits, so it can't reproduce this. The full engine TestRunner passes: 375/375 doctest cases, and no JUCE test failures.

App PR: Tracktion/waveform#135


drowaudio:
cleanUpDanglingPlugins seems a bit heavyweight as it delete the AsyncPluginDeleter. It's probably better to just expose a function that internally calls what the timerCallback does to flush the list. Make sure to add a message-thread assert in that function

drowaudio:
🤖 Done in 458a938. Added `deleteQueuedPlugins()` next to `cleanUpDanglingPlugins()` in `tracktion_ExternalPlugin.cpp`. It has `TRACKTION_ASSERT_MESSAGE_THREAD` and calls a new `AsyncPluginDeleter::deleteQueuedPlugins()`, which runs the same step as `timerCallback()` until the queue is empty (stopping if a deletion re-enters). The singleton stays alive, so the timer and any `waitForPluginDeletion()` callback still work. `convertToFolderBasedProject` now calls it, still only on the message thread. Engine TestRunner: 375/375. Waveform's sampler tests, including the conversion one, all pass.

